PR #24, opened 2026-01-10
Adding github action for static checks

clayrosenthal:
*Description of changes:*

Adding github action to run static checks in the holosoma container

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -1,0 +1,64 @@
+name: Static Checks
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  pull_request:
+    types:
+    - opened
+    - reopened
+    - synchronize
+
+jobs:
+  login-to-amazon-ecr:
+    name: Setup for ECR
+    runs-on:
+      # - codebuild-holosoma-gpu-build-${{ github.run_id }}-${{ github.run_attempt }}
+      - codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-west-2
+          role-session-name: gha-${{ github.run_id }}-${{ github.run_attempt }}
+      - name: Setup ecr-cred helper
+        run: |
+          sudo apt update
+          sudo apt install -y amazon-ecr-credential-helper
+          mkdir -p ~/.docker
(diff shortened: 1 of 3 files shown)


.github/workflows/static-checks.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - opened
+    - reopened
+    - synchronize
+
+jobs:
+  login-to-amazon-ecr:
+    name: Setup for ECR
+    runs-on:
+      - codebuild-holosoma-gpu-build-${{ github.run_id }}-${{ github.run_attempt }}
+      # - codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
```
lindashih: should we run this on cpu node instead of gpu?

clayrosenthal: Yeah, very reasonable, I'll need to double check the cpu runner's AMI, didn't work when I tried it earlier

.github/workflows/static-checks.yaml
```diff
@@ -0,0 +1,64 @@
...
+          source scripts/source_isaacgym_setup.sh
+          pip install pre-commit mypy
+          pip install -e src/holosoma
+          pip install -e src/holosoma_inference
+          git fetch origin "${GITHUB_BASE_REF:-main}"
+          git fetch origin "${GITHUB_SHA}"
+          pre-commit install-hooks
+          git diff --name-only "origin/${GITHUB_BASE_REF:-main}" "${GITHUB_SHA}" --
+          git diff --name-only "origin/${GITHUB_BASE_REF:-main}" "${GITHUB_SHA}" -- | xargs pre-commit run --files
+          mypy .
```
samuelgundry, line 64: @clayrosenthal You shouldn't need to manually run pip install on holosoma. Hopefully you can run the setup scripts only, that setup the dependencies and conda env?  If you need both inference and core installed, perhaps see if we can make the setup scripts work together (use same conda env?) as I think installing in the same environment might be a valid use case outside CI. E

E.g, something like: 
 ```
bash scripts/setup_isaacgym.sh my_hsenv
bash scripts/setup_inference.sh my_hsenv
```

Ping me though if this seems unnecessary. My main worry is making local dev and CI checks consistent.

clayrosenthal: Yeah I'm open to that. I can try that and see how we can shrink this script

lindashih: let's do those improvements as a follow-up to getting the actions running though

PR #27, opened 2026-01-13
Adding nightly training

clayrosenthal:
*Description of changes:*

Adding robot sim nightly training for G1 29-DOF, T1. Runs on issacsim and isaacgym, as well as multi gpu and fast-sac. WBT will be added in a follow up


clayrosenthal:
Last run was mostly successful, but one timed out, [re-running tests with increased timeout](https://github.com/amazon-far/holosoma/actions/runs/21078125949)

PR #38, opened 2026-01-30
Updating nightly gha to always post to slack, use aws creds for s3 access

clayrosenthal:
*Description of changes:*

Reordering steps in evaluate to always post the slack message. Tagging the runs with the gha run ID and filtering for that when reading the summary. 

Also passing AWS creds so we can pull from S3 properly


clayrosenthal:
[small test run for wbt](https://github.com/amazon-far/holosoma/actions/runs/21498026472)

and [full nightly suite](https://github.com/amazon-far/holosoma/actions/runs/21500219395)

PR #39, opened 2026-01-30
pruning old nightly docker containers, pass env vars

clayrosenthal:
*Description of changes:*

Fixing docker runs running out of memory because of old containers, and making sure env vars are passed correctly so run tags are set properly


clayrosenthal:
testing on [manual nightly run](https://github.com/amazon-far/holosoma/actions/runs/21605121375)

.github/workflows/nightly-training.yaml
```diff
@@ -48,13 +48,21 @@ jobs:
         env:
           WANDB_API_KEY: ${{ secrets.WANDB_PUB_API_KEY }}
         shell: bash
-        timeout-minutes: 750 # 12.5 hours
+        timeout-minutes: 1000 # ~16 hours
+
         run: |
+          # kill all old docker containers
```
samuelgundry, line 54: we'll eventually want multi runs on the same machine (for availability), so just noting we'll need to revisit this later

PR #41, opened 2026-02-02
Add OSX support for mujoco training and evaluation

gabriellemerritt:
# Summary
Added macOS support to the Holosoma robotics framework by:

Detecting OS/architecture and installing appropriate Miniconda version
Conditionally installing Linux-only dependencies
Adding macOS-specific installation and usage instructions to README
Updating platform tag detection for proper wheel installation
Upgrading wandb to v0.23+ to support longer API keys

# OSX Performance Benchmark

**Platform:** macOS (Darwin)  
**Simulator:** MuJoCo CPU (Classic backend)  
**Date:** January 30, 2026  
**Configuration:** 100 iterations, 1 environment, seed 42

## Performance Results

| Robot | Algorithm | **Eval Time** | **Training Time** | **Time/Iter** | **Eval Steps/s** | **Train Iter/s** |
|-------|-----------|---------------|-------------------|---------------|------------------|------------------|
| G1 (29-DOF) | PPO | 16s | 71s | 0.71s | 62.5 | 1.41 |
| G1 (29-DOF) | FastSAC | 16s | 313s | 3.13s | 62.5 | 0.32 |

### Evaluation Performance
- Both algorithms take exactly **16 seconds** for evaluation (1000 steps)
- Identical throughput: **62.5 steps/second** regardless of algorithm
- Algorithm choice has **zero impact** on inference speed

### Training Performance
- **PPO is 4.4x faster** than FastSAC (71s vs 313s)
- PPO achieves **1.41 iterations/second** vs FastSAC's **0.32 iter/s**
- PPO completes 100 iterations in just over 1 minute vs FastSAC's 5+ minutes

### CPU Utilization
- Extremely low: **0.001-0.01% average**, **0.4-1.1% peak**
- Training is simulation-bound, not compute-bound
- Single environment bottleneck confirmed

## Configuration Details

**Test Configuration:**
- Iterations: 100
- Environments: 1 (Single environments only for CPU MuJoCo)
- Seed: 42
- Simulator: MuJoCo Classic backend
- Python: mjpython interpreter

**Hardware:**
- Platform: macOS (Darwin)
- CPU utilization: 0.001-0.01% average, 0.4-1.1% peak
- Memory utilization: 0% (negligible)

### DIFF
--- a/src/holosoma_inference/setup.py
+++ b/src/holosoma_inference/setup.py
@@ -1,23 +1,38 @@
 import platform
+import sys
 
 from setuptools import find_packages, setup
 
-UNITREE_VERSION = "0.1.1"
+UNITREE_VERSION = "0.1.2"
 UNITREE_REPO = "https://github.com/amazon-far/unitree_sdk2"
 BOOSTER_VERSION = "0.1.0"
 BOOSTER_REPO = "https://github.com/amazon-far/booster_robotics_sdk"
 
-PLATFORM_MAP = {
-    "x86_64": "linux_x86_64",
-    "aarch64": "linux_aarch64",
-}
 
-platform_tag = PLATFORM_MAP.get(platform.machine(), "linux_x86_64")
+def get_platform_tag():
+    """Get the correct platform tag for wheel installation."""
+    machine = platform.machine()
+    system = platform.system()
 
+    if system == "Linux":
+        if machine == "x86_64":
+            return "linux_x86_64"
+        if machine in ("aarch64", "arm64"):
+            return "linux_aarch64"
+    elif system == "Darwin":
+        # Get macOS version for proper wheel tag
+        mac_ver = platform.mac_ver()[0]  # Returns version like "14.1.0"
+        major_version = mac_ver.split(".")[0] if mac_ver else "11"
+        assert major_version == "26"
+        return f"macosx_{major_version}_0_{machine}"
+    # Fallback
+    return "linux_x86_64"
+
+
+platform_tag = get_platform_tag()
+pyvers = f"cp{sys.version_info.major}{sys.version_info.minor}"
 unitree_extras = []
-unitree_url = (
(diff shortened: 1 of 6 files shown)


src/holosoma_inference/setup.py
```diff
@@ -1,23 +1,38 @@
 import platform
+import sys
 
 from setuptools import find_packages, setup
 
-UNITREE_VERSION = "0.1.1"
+UNITREE_VERSION = "0.1.2" if platform.machine() == "arm64" else "0.1.1"
```
samuelgundry: I'd prefer the same version across platforms if possible.  

Is this just to reduce testing or is there a technical issue?  I'm okay either way, just for my knowledge, and I realize testing takes time.

src/holosoma_inference/setup.py
```diff
@@ -1,23 +1,38 @@
...
 BOOSTER_VERSION = "0.1.0"
 BOOSTER_REPO = "https://github.com/amazon-far/booster_robotics_sdk"
 
-PLATFORM_MAP = {
-    "x86_64": "linux_x86_64",
-    "aarch64": "linux_aarch64",
-}
 
-platform_tag = PLATFORM_MAP.get(platform.machine(), "linux_x86_64")
+def get_platform_tag():
```
samuelgundry, line 12: We would benefit figuring a clean way to share modules. We separated hsinference intentionally to reduce dependencies/heaviness but given these are in the same repo, a helper/library package would be nice in a few places. Not for this PR.

samuelgundry (approved):
Looks good. Minor question w.r.t dep versioning

src/holosoma_inference/setup.py
gabriellemerritt: Looking at the history, I think this should be 0.1.2 for all platforms. In this [CR](https://github.com/amazon-far/holosoma/commit/d8b32663b25962a4af708a519a36ec03ed3eeb5d) it looks like `src/holosom_inference/setup.py` was missed

Issue #44, opened 2026-02-05
Training creates two experiment directories with different timestamps


maedmatt:
Hello, 

I just encountered an issue inside the training script.

Every training run creates two sibling experiment directories under `logs/`:

  logs/project/XXX/  ← weights, config, etc.
  logs/project/XXX/  ← renderings_training/ (empty)

 `get_timestamp()` is called twice independently:
  1. `train_agent.py:182` — generates the timestamp used for logging, config, checkpoints
  2. `base_task.py:82` — generates a second timestamp a few seconds later, used for the simulator/video recorder's `experiment_dir`

Since the environment setup (IsaacGym init, wandb init, etc.) takes a few seconds between the two calls, the timestamps differ, creating two directories.

The video recorder writes to the second directory (`renderings_training/`), but when wandb is enabled the local video files are cleaned up after upload, leaving it empty.

I have created a commit that should solve the issue, should I open a PR? 

Thanks for the help!

PR #46, opened 2026-02-06
fix double experiment directory creation during training

maedmatt:
## Issue
Issue is described in #44

## Fix
Pass the `experiment_dir` computed in `train_agent.py` through `EnvConfig` to `base_task.py` instead of having `base_task.py` compute its own. Added an optional `experiment_dir: str | None = None` field to `EnvConfig`. Other callers (replay, eval, tests) that don't pass it get the existing fallback behavior.

## Test plan

   - [x] `ruff check` passes
   - [x] `mypy` passes (no new issues)
   - [x] Full test suite passes (151 passed, 2 skipped)
   - [x] Smoke test: `python -m holosoma.train_agent exp:g1-23dof --algo.config.num_learning_iterations=5` creates exactly one directory
   
### DIFF
--- a/src/holosoma/holosoma/config_types/env.py
+++ b/src/holosoma/holosoma/config_types/env.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from pydantic.dataclasses import dataclass
+
 from holosoma.config_types.action import ActionManagerCfg
 from holosoma.config_types.command import CommandManagerCfg
 from holosoma.config_types.curriculum import CurriculumManagerCfg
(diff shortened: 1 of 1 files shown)


src/holosoma/holosoma/envs/base_task/base_task.py
```diff
@@ -76,13 +77,17 @@ def __init__(
...
-        timestamp = get_timestamp()
-        experiment_dir = get_experiment_dir(
-            tyro_config.logger, tyro_config.training, timestamp, task_name=self._get_task_name()
-        )
+        # Use pre-computed experiment directory if provided (from train_agent.py),
+        # otherwise compute one (for replay, eval, etc.)
+        if tyro_config.experiment_dir is not None:
+            experiment_dir = Path(tyro_config.experiment_dir)
+        else:
+            from holosoma.utils.experiment_paths import get_experiment_dir, get_timestamp  # noqa: PLC0415
```
clayrosenthal: Could you import this at the top of the file? Looks like we didn't before, but unsure why we wouldn't.

maedmatt: I've moved the top-level import in 34c243a

PR #49, opened 2026-02-10
Add option to run part of WBT motion clip + refactor timestepping code

tomasz-lewicki:
## Description

This PR adds an option to play just a part of the WBT motion clip.

Additionally:
- Refactored the inline simulation-time code to two single-purpose utility classes: `MotionClockUtil`, `TimestepUtil`
- Added  `wbt_utils.py` to separate the policy implementation from these utils.  
- Added `test_wbt_utils.py` + github workflow to [Run Inference Tests (pull_request)](https://github.com/amazon-far/holosoma/actions/runs/21884814906/job/63176850472?pr=49)
## Testing

Sim:
```bash
source scripts/source_mujoco_setup.sh 
python src/holosoma/holosoma/run_sim.py robot:g1-29dof simulator:mujoco
```
Policy:
```bash
python3 src/holosoma_inference/holosoma_inference/run_policy.py inference:g1-29dof-wbt     --task.model-path src/holosoma_inference/holosoma_inference/models/wbt/fastsac_g1_29dof_dancing.onnx     --task.no-use-joystick     --task.rl-rate 50     --task.interface lo     --robot.sdk-type unitree     --task.use-sim-time       --task.motion-start-timestep 1500     --task.motion-end-timestep 1600
```

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -38,11 +38,12 @@ jobs:
           bash tests/ci/${{ matrix.simulator }}_ci_tests.sh
 
   inference-tests:
-    name: Run Inference Tests
+    name: Run Inference Unit Tests
     runs-on: codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
     container:
       # TODO: Switch to holosoma-inference:latest once it's available in ECR
       # image: 982423663241.dkr.ecr.us-west-2.amazonaws.com/holosoma-inference:latest
+      # Until then, the inference tests will run in isaacsim env.
       image: 982423663241.dkr.ecr.us-west-2.amazonaws.com/holosoma:latest
     steps:
       - name: Checkout code


src/holosoma_inference/holosoma_inference/policies/wbt.py
```diff
@@ -295,13 +246,12 @@ def get_current_obs_buffer_dict(self, robot_state_data):
     def rl_inference(self, robot_state_data):
         # prepare obs, run policy inference
         if not self.motion_clip_progressing:
-            # Keep motion index pinned at the start while waiting to trigger the clip.
-            self.motion_timestep = 0
-            self.motion_start_timestep = None
-            self._last_clock_reading = None
+            # Keep motion index pinned at the configured start while waiting to trigger the clip.
```
carlosferrazza, line 249: Have you checked that the behavior is the same at timestep 0?

carlosferrazza (approved):

src/holosoma_inference/holosoma_inference/policies/wbt_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        root = ElementTree.fromstring(urdf_text)
+
+        def _is_visual_or_collision(tag: str) -> bool:
+            # Handle optional XML namespaces by only checking the suffix after '}'.
+            return tag.split("}")[-1] in {"visual", "collision"}
+
+        for parent in root.iter():
+            for child in list(parent):
+                if _is_visual_or_collision(child.tag):
+                    parent.remove(child)
```
clayrosenthal, line 133: removing elements from a list while iterating in it can cause issues, try using list comprehension:
```
parent = [ child for child in list(parent) if not _is_visual_or_collision(child.tag) ]
```

clayrosenthal: tho might be different for this xml library, not sure

.github/workflows/pytest.yaml
```diff
@@ -36,3 +36,26 @@ jobs:
             ln -sfF "$GITHUB_WORKSPACE" /workspace/holosoma
           fi
           bash tests/ci/${{ matrix.simulator }}_ci_tests.sh
+
+  inference-tests:
+    name: Run Inference Unit Tests
+    runs-on: codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
```
clayrosenthal, line 42: does this not need GPUs?

clayrosenthal (approved):

src/holosoma_inference/holosoma_inference/policies/wbt_utils.py
tomasz-lewicki: I think we're iterating over the copy (i.e. `list(parent)`) and calling `.remove()` on the original `parent` so we should be fine.

.github/workflows/pytest.yaml
tomasz-lewicki: CPU only for now. I figured I'd move it to GPU once we actually have tests that use GPUs.

src/holosoma_inference/holosoma_inference/policies/wbt.py
tomasz-lewicki: Yes, same behavior:
1. Happy path looks the same
2. Clock jump-back behavior (e.g. we click `backspace` in mujoco during execution) looks the same.
3. Clock jump-ahead behavior (e.g. start the policy `]`, but don't start the motion `s` for a while) looks the same.

PR #50, opened 2026-02-12
Add an option to print observations in WBT

tomasz-lewicki:
Minor PR to optionally print the observation vector in `wbt.py` when `--task.print-observations` is passed.


Juyue (approved):
lgtm!

PR #51, opened 2026-02-13
Publish wandb reports, update nightly runs

clayrosenthal:
*Description of changes:*

Adds a step to the nightly workflow to create and publish a wandb report for each set of nightly runs. Also updating the nightly runs to all run in the same project with better run names/ids. 

[Small test run](https://github.com/amazon-far/holosoma/actions/runs/21966061114)
[Full test run](https://github.com/amazon-far/holosoma/actions/runs/21967641227)

### DIFF
--- a/tests/nightly/generate_report.py
+++ b/tests/nightly/generate_report.py
@@ -1,0 +1,376 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "wandb",
+#   "wandb-workspaces",
+# ]
+# ///
+
+from __future__ import annotations
+
+import argparse
+from datetime import datetime, timezone
+from os import getenv
+
+import wandb
+import wandb_workspaces.reports.v2 as wr
+
+WANDB_ENTITY = getenv("WANDB_ENTITY", "amazon-far")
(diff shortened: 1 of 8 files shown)


tests/nightly/generate_report.py
```diff
@@ -0,0 +1,376 @@
...
+    """
+    blocks = [
+        wr.H1(f"Experiment: {exp_name}"),
+        wr.P(f"Analysis of {len(run_ids)} nightly runs for {exp_name}"),
+    ]
+
+    # Define metric groups based on experiment type
+    if "wbt" in exp_name:
+        # Whole Body Tracking metrics
+        metric_groups = {
```
samuelgundry, line 133: we'd probably eventually want to pull these from dedicate config and/or existing configs rather than in the report builder logic, in case names change, so scientists can maintain the report content, but all good for now!

samuelgundry (approved):
Nice work. Great to have reports and nightly runs.

tests/nightly/generate_report.py
clayrosenthal: good idea. When I had claude cook this up it didn't know a lot of the metric names. Would be best to have one source of truth

PR #52, opened 2026-02-13
Inference improvements

tomasz-lewicki:
## Misc Inference code improvements

1. Auto-detect network interface: `TaskConfig.interface` defaults to `"auto"` instead of `"lo"`. When `"auto"`, a new `detect_robot_interface()` function scans `/sys/class/net/` for the first wired NIC that is operationally UP, skipping wifi/docker/veth interfaces. If no interface is found, we fallback to `lo` 

2. Added debug config overrides for easier testing. Specifically `DebugConfig` to `TaskConfig` with three toggles: `force_upright_imu` (override gravity to `[0,0,-1]`), `force_zero_angular_velocity`, and `force_zero_action` (hold default pose). Also added a per-joint diagnostic table (`_print_joint_table`) that prints q/dq/action in degrees alongside observation printing. This is super helpful for debugging policies.

3. Freejoint detection in MuJoCo: The simulator no longer looks for a hardcoded `"floating_base_joint"` name. Instead it finds the freejoint by `mjJNT_FREE` type, and filters robot joints by joint type rather than name exclusion. This fixes compatibility with robot XMLs that don't follow this naming convention.

4. Lazy extension entrypoint loading: `holosoma.config.inference` and `holosoma.config.robot` entry points are now loaded on first use (`_load_extensions()`) instead of at module import time. This helps us avoid circular imports in extensions, while preseriving the nice `--help` messages in core and extensions. The `AnnotatedInferenceConfig` is replaced by a `get_annotated_inference_config()` factory that triggers lazy loading before building the tyro CLI type.

5. Suppress problematic tyro CLI fields — `SceneConfig.scene_files` and `SceneConfig.rigid_objects` are annotated with `tyro.conf.Suppress` to prevent opaque parser errors when users pass unrecognized CLI flags. Previously, we would somehow always get error about these fields whenever _any_ unrecognized field would get in.

6. Remove dead SDK code. Deleted `sdk/command_sender/` and `sdk/state_processor/`. These were superseded by the entry-point-based `sdk/__init__.py` interface factory and are no longer imported anywhere. I should've deleted them in #45.

7. Added python3.12 mujoco setup scripts. I believe @kingb has a cleaner solution, though.

8. Docker Improvements: Inference docker now mounts `/dev/input` (for joystick) and persists bash history across container restarts. Added `evdev` to `setup.py` dependencies.

9. WBT observation printing: the `WholeBodyTrackingPolicy` now respects `print_observations` flag (was previously only wired in locomotion).

10. Remove unused upper controller path in base.py policy postprocessing.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -84,10 +84,9 @@ jobs:
     name: Run Inference Unit Tests
     runs-on: codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
     container:
-      # Uses holosoma:latest which has all conda envs (hsgym, hssim, hsinference, etc.)
-      # built via setup_all.sh. We activate hsinference (Python 3.10) for inference tests.
       # TODO: Switch to `holosoma-inference` once it's been confirmed functional
--     # image: 982423663241.dkr.ecr.us-west-2.amazonaws.com/holosoma-inference:latest
+      # image: 982423663241.dkr.ecr.us-west-2.amazonaws.com/holosoma-inference:latest
+      # Until then, the inference tests will run in isaacsim env.
       image: 982423663241.dkr.ecr.us-west-2.amazonaws.com/holosoma:latest
     steps:
       - name: Checkout code
@@ -100,6 +99,6 @@ jobs:
             rm -rf /workspace/holosoma
             ln -sfF "$GITHUB_WORKSPACE" /workspace/holosoma
           fi
-          source scripts/source_inference_setup.sh
+          source scripts/source_isaacsim_setup.sh
           pip install -e src/holosoma_inference
           pytest -s src/holosoma_inference/


src/holosoma_inference/docker/run.sh
```diff
@@ -17,6 +17,11 @@ PROJECT_ROOT="$( cd "$SCRIPT_DIR/../../.." && pwd )"
 CONTAINER_NAME="holosoma-inference-container"
 IMAGE_NAME="holosoma-inference"
 
+# Mount bash history in the host filesystem in order to preserve history between containers
+HOLOSOMA_DEPS_DIR="$HOME/.holosoma_deps"
+mkdir -p "$HOLOSOMA_DEPS_DIR"
+touch "$HOLOSOMA_DEPS_DIR/bash_history"
```
clayrosenthal, line 23: should we separate this per container/env? Like if I ran something in the inference container, do I necessarily also want that history in the retargeting or isaacsim container?

src/holosoma_inference/holosoma_inference/run_policy.py
clayrosenthal, line 1: might be out of scope for this PR, but should we actually spell out the [package entrypoints](https://setuptools.pypa.io/en/latest/userguide/entry_point.html) in `pyproject.toml` or `setup.py`?

src/holosoma/holosoma/simulator/mujoco/mujoco.py
```diff
@@ -456,21 +459,19 @@ def _set_robot_joint_addressing(self) -> None:
...
         if self.robot_freejoint_id == -1:
-            logger.warning(f"Robot freejoint '{freejoint_name}' not found in model")
+            logger.warning("No freejoint found in model, using joint 0 as fallback")
             self.robot_freejoint_id = 0
-            has_freejoint = False
-
-        # Validate it's actually a freejoint
-        if has_freejoint and self.root_model.jnt_type[self.robot_freejoint_id] != mujoco.mjtJoint.mjJNT_FREE:
-            joint_type = self.root_model.jnt_type[self.robot_freejoint_id]
-            raise ValueError(f"Joint '{freejoint_name}' is not a freejoint, got type {joint_type}")
```
samuelgundry, line 473: Changes look good. I wish I could remember why I didn't use the joint type originally as the filter. It might have just been a miss on my part -- if these work for our supported robot types, I'm good.

scripts/setup_mujoco_py312.sh
```diff
@@ -0,0 +1,71 @@
...
+pip install "mujoco>=3.0.0" mujoco-python-viewer
+pip install torch torchvision --index-url https://download.pytorch.org/whl/cu128
+
+# --- Holosoma runtime deps (compatible with Python 3.12) ---
+pip install astor easydict ipdb joblib loguru lxml matplotlib meshcat omegaconf \
+    opencv-python plotly pygame pynput rich scipy tensorboard tensordict \
+    termcolor tqdm trimesh "yourdfpy>=0.0.58" zmq shapely click \
+    "warp-lang>=1.10" pydantic "tyro>=1.0.0" "numpy<2"
+
+# --- Install holosoma + extensions (--no-deps to skip numpy==1.23.5 pin) ---
```
clayrosenthal, line 53: There was a discussion in the uv PR (#70) to unpin numpy, do you know why we pinned it in the first place? I assume later numpy versions have a wheel for 3.12

src/holosoma_inference/holosoma_inference/utils/network.py
```diff
@@ -0,0 +1,21 @@
...
+            continue
+        try:
+            with open(f"/sys/class/net/{ifname}/operstate") as f:
+                if f.read().strip().lower() != "up":
+                    continue
+        except OSError:
+            continue
+        print(f"[network] auto-detected interface: {ifname}")
+        return ifname
+    raise RuntimeError("No wired NIC found. Specify --interface explicitly.")
```
lindashih: I don't see the fallback to `lo` here

src/holosoma_inference/holosoma_inference/utils/network.py
```diff
@@ -0,0 +1,21 @@
+"""Network interface auto-detection for robot communication."""
+
+import os
+
+_SKIP_PREFIXES = ("lo", "wl", "docker", "br-", "veth", "virbr", "vnet", "tun", "tap")
+
+
+def detect_robot_interface(robot_type: str | None = None) -> str:
```
lindashih: where is `robot_type` getting used?

scripts/source_mujoco_py312_setup.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# Source this file to activate the hsmujoco_py312 conda env (Python 3.12 + ROS2 Jazzy).
+# Usage:  source holosoma/scripts/source_mujoco_py312_setup.sh
+
+# Detect script directory (works in both bash and zsh)
+if [ -n "${BASH_SOURCE[0]}" ]; then
+    SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+elif [ -n "${ZSH_VERSION}" ]; then
+    SCRIPT_DIR=$( cd -- "$( dirname -- "${(%):-%x}" )" &> /dev/null && pwd )
+fi
```
asetapen: I like how both cases are explicit here! I think this will break if symlinked. Here is my tried-and-true `bash` and `zsh` and `docker` supported way to do this:

```bash
SOURCE="${BASH_SOURCE[0]:-${(%):-%x}}"
while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null && pwd )"
  SOURCE="$(readlink "$SOURCE")"
  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
done
SCRIPT_DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null && pwd )"
```

src/holosoma_inference/holosoma_inference/policies/base.py
```diff
@@ -451,8 +451,35 @@ def _print_observations(self, obs: dict[str, np.ndarray]) -> None:
                     term_values = group_obs[0, start_idx : start_idx + total_dim]
                     print(f"  {term_name:20s} (dim={term_dim:2d}, hist={history_len}): {term_values}")
                     start_idx += total_dim
+
+        # Joint table: dof_name | q (deg) | dq | action
+        self._print_joint_table(obs)
         print("========================================\n")
 
+    def _print_joint_table(self, obs: dict[str, np.ndarray]) -> None:
```
asetapen, line 470: super useful - thanks for adding this!

scripts/setup_mujoco_py312.sh
tomasz-lewicki: I think we wanted to just pin it to `<2` for binary compatibility and someone got over-eager and pinned it to a specific version. Let me see if I can relax it in a future PR.

src/holosoma_inference/holosoma_inference/utils/network.py
tomasz-lewicki: I haven't pushed the change before review!
Added in 7e1e350b5aedb67259c8700dbc85e9136d3b3301

src/holosoma_inference/holosoma_inference/utils/network.py
tomasz-lewicki: It's not, thanks for pointing this out.
I had more complicated logic based on `robot_type` but decided to simplify.

tomasz-lewicki: Fixed in 440b4f4

src/holosoma_inference/holosoma_inference/run_policy.py
tomasz-lewicki: I think they are explicitly listed in:
 https://github.com/amazon-far/holosoma/blob/440b4f49c0d508c6ab26f736072a81d15e5ad735/src/holosoma_inference/setup.py#L64

Is this what you had in mind?

scripts/source_mujoco_py312_setup.sh
tomasz-lewicki: Added in 8bb51af

src/holosoma_inference/holosoma_inference/run_policy.py
tomasz-lewicki: Added in [f659ce5](https://github.com/amazon-far/holosoma/pull/52/commits/f659ce5f32a95f90c5527ad5fb376bf2297c8ac9)

PR #58, opened 2026-02-19
Docker image per environment

clayrosenthal:
*Description of changes:*

Making more docker images for each env to have smaller images for users who only want one part of holosoma.

This is just the first step, ideally the base image can be replaced for all other non-isaacsim images.

New images:
| Image Name | Disk Usage | Content Size |
|--------|--------|--------|
| holosoma-inference | 3.11GB | 819MB |
| holosoma-isaacsim | 68GB | 22.8GB |
| holosoma-isaacgym | 50.1GB | 17.3GB | 
| holosoma-mujoco | 54.8GB | 19.5GB |
| holosoma-retargeting | 48.8GB | 17.5GB |

vs old image:
| Image Name | Disk Usage | Content Size |
|--------|--------|--------|
| holosoma  | 132GB | 47.3GB |


docker/mujoco.Dockerfile
```diff
@@ -0,0 +1,59 @@
+# MuJoCo-only image
+# Runs setup_mujoco.sh to create the hsmujoco conda environment (Python 3.10)
+# with MuJoCo >= 3.0.0 and holosoma installed.
+#
+# GPU/Warp acceleration is disabled (--no-warp) for image build compatibility.
+# To enable GPU-accelerated MuJoCo Warp, rebuild without the --no-warp flag:
+#   docker build --build-arg WARP=true -f docker/mujoco.Dockerfile ...
+# and ensure NVIDIA driver >= 550.54.14 is present on the host.
+FROM nvcr.io/nvidia/isaac-sim:5.1.0
```
samuelgundry, line 9: What's the reason for using this big Isaac image for Mujoco?   Related, did we want a cpu-only and GPU/warp Mujoco image?

samuelgundry (commented):
I see it's merged, leaving a comment for future consideration.

docker/mujoco.Dockerfile
clayrosenthal: I was trying to minimize changes. I mention in the description we could probably switch most of these to smaller base images, this was just was just to start with a version that works to compare against as we pare down to the essentials. I think a CPU variant would make sense as well to really slim down the size

PR #71, opened 2026-03-09
Add uv-based MuJoCo setup scripts as alternative to conda

ghost:
## Summary

- Adds `scripts/setup_mujoco_via_uv.sh` — a lighter-weight alternative to the conda-based `setup_mujoco.sh`, using uv + standard venvs instead of conda/mamba/miniconda
- Adds `scripts/source_mujoco_uv_setup.sh` — activation script for the uv-based environment (no `source_common.sh` dependency)
- Loosens numpy pin from `==1.23.5` to `>=1.23.5` for Python 3.12 compatibility
- Updates NVIDIA driver install example from 550 to 590 in both setup scripts

### New flags (uv script)
- `--no-warp` — skip GPU/Warp installation (same as conda version)
- `--no-robot-sdks` — skip unitree/booster SDK installation
- `--python VERSION` — specify Python version for the venv

### Notes
- Robot SDK wheels (unitree, booster) are Linux-only; the script detects macOS and skips them gracefully
- Tested on macOS arm64 with Python 3.12

## Test plan
- [x] Tested `setup_mujoco_via_uv.sh --no-warp` on macOS arm64 / Python 3.12
- [x] Tested `source_mujoco_uv_setup.sh` activation and validation
- [ ] Test on Ubuntu amd64 with GPU (warp path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/setup_mujoco_via_uv.sh
+++ b/scripts/setup_mujoco_via_uv.sh
@@ -1,0 +1,208 @@
+#!/bin/bash
+# Exit on error, and print commands
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$(dirname "$SCRIPT_DIR")
+
+# MuJoCo Warp version to install -- the repo is missing version tags and branches
+# Arbitrarily chosen from mainline at the time we've ~tested against
+MUJOCO_WARP_COMMIT="09ec1da"
+
+# Workspace and venv configuration
+WORKSPACE_DIR=$HOME/.holosoma_deps
+VENV_DIR=$WORKSPACE_DIR/venvs/hsmujoco
+
+# Parse command-line arguments
+INSTALL_WARP=true
+INSTALL_ROBOT_SDKS=true
+PYTHON_VERSION=""
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --no-warp)
+      INSTALL_WARP=false
+      echo "MuJoCo Warp (GPU) installation disabled - CPU-only mode"
+      shift
+      ;;
+    --no-robot-sdks)
+      INSTALL_ROBOT_SDKS=false
+      echo "Robot SDK installation disabled (unitree, booster)"
+      shift
+      ;;
+    --python)
+      PYTHON_VERSION="$2"
+      shift 2
+      ;;
+    --help|-h)
+      echo "Usage: $0 [--no-warp] [--no-robot-sdks] [--python VERSION]"
+      echo ""
+      echo "Options:"
+      echo "  --no-warp          Skip MuJoCo Warp installation (CPU-only)"
+      echo "  --no-robot-sdks    Skip robot SDK installation (unitree, booster)"
+      echo "  --python VERSION   Python version to use (e.g., 3.10, 3.11)"
+      echo "  --help, -h         Show this help message"
+      echo ""
+      echo "Default: GPU-accelerated installation with robot SDKs"
+      echo ""
+      echo "Examples:"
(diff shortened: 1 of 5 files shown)


scripts/setup_mujoco_via_uv.sh
```diff
@@ -0,0 +1,208 @@
...
+    echo "Reference: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/"
+    echo ""
+    echo "After driver installation, re-run this script"
+    echo "(or use --no-warp for CPU-only installation)"
+    exit 1
+  fi
+
+  echo "NVIDIA driver version: $DRIVER_VERSION (meets minimum $MIN_DRIVER_VERSION)"
+
+  if [[ ! -d $WORKSPACE_DIR/mujoco_warp ]]; then
```
clayrosenthal, line 177: why not use `uv pip install "mujoco-warp @ git+https://github.com/google-deepmind/mujoco_warp.git@${MUJOCO_WARP_COMMIT}"` or just `uv pip install mujoco-warp`? It's on [pypi](https://pypi.org/project/mujoco-warp)

scripts/setup_mujoco_via_uv.sh
```diff
@@ -0,0 +1,208 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$(dirname "$SCRIPT_DIR")
+
+# MuJoCo Warp version to install -- the repo is missing version tags and branches
+# Arbitrarily chosen from mainline at the time we've ~tested against
+MUJOCO_WARP_COMMIT="09ec1da"
+
+# Workspace and venv configuration
+WORKSPACE_DIR=$HOME/.holosoma_deps
```
clayrosenthal, line 13: Given we're not installing non-python packages for this setup, I'm not sure we really need to follow this weird pattern of a specific folder to install, I feel like this should be in the repo directory like most other python venvs. Will probably still make sense to have more than one venv, so like `.venv/hsmujoco`.

scripts/setup_mujoco_via_uv.sh
```diff
@@ -0,0 +1,208 @@
...
+      echo "Unknown option: $1"
+      echo "Usage: $0 [--no-warp] [--no-robot-sdks] [--python VERSION]"
+      echo "Use --help for more information"
+      exit 1
+      ;;
+  esac
+done
+
+# Sentinel files
+SENTINEL_FILE=${WORKSPACE_DIR}/.env_uv_setup_finished_hsmujoco
```
clayrosenthal, line 71: Also don't feel like this sentinel file thing is needed, I feel like we could just run `uv sync` with (possibly [conflicting](https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies)) [groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups), tho maybe that can be a follow up

scripts/source_mujoco_uv_setup.sh
```diff
@@ -0,0 +1,42 @@
...
+if [[ ! -d "$VENV_DIR" ]]; then
+    echo "Error: uv MuJoCo environment not found at $VENV_DIR"
+    echo "Run 'bash scripts/setup_mujoco_via_uv.sh' first."
+    return 1 2>/dev/null || exit 1
+fi
+
+source "$VENV_DIR/bin/activate"
+
+# Set MuJoCo-specific environment variables
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}$VENV_DIR/lib
```
clayrosenthal, line 24: Is this needed or is this a copy of how conda was doing it?

PR #72, opened 2026-03-10
Add uv-based MuJoCo setup scripts as alternative to conda

kingb:
## Summary

- Adds `scripts/setup_mujoco_via_uv.sh` — a lighter-weight alternative to the conda-based `setup_mujoco.sh`, using uv + standard venvs instead of conda/mamba/miniconda
- Adds `scripts/source_mujoco_uv_setup.sh` — activation script for the uv-based environment (no `source_common.sh` dependency)
- Loosens numpy pin from `==1.23.5` to `>=1.23.5` for Python 3.12 compatibility
- Updates NVIDIA driver install example from 550 to 590 in both setup scripts

### New flags (uv script)
- `--no-warp` — skip GPU/Warp installation (same as conda version)
- `--no-robot-sdks` — skip unitree/booster SDK installation
- `--python VERSION` — specify Python version for the venv

### Notes
- Robot SDK wheels (unitree, booster) are Linux-only; the script detects macOS and skips them gracefully
- Tested on macOS arm64 with Python 3.12

## Test plan
- [x] Tested `setup_mujoco_via_uv.sh --no-warp` on macOS arm64 / Python 3.12
- [x] Tested `source_mujoco_uv_setup.sh` activation and validation
- [ ] Test on Ubuntu amd64 with GPU (warp path)

### DIFF
--- a/src/holosoma/pyproject.toml
+++ b/src/holosoma/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     "mypy==1.14.1",
     "notebook",
     "numpy-stl",
-    "numpy>=1.23.5",
+    "numpy>=1.23.5,<2",
     "omegaconf",  # Used for config resolvers and DictConfig type hints
     "onnx",
     "onnxruntime",


src/holosoma/pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
     "mypy==1.14.1",
     "notebook",
     "numpy-stl",
-    "numpy==1.23.5",
+    "numpy>=1.23.5",
```
tomasz-lewicki: We should also add `<2` for binary compatibility.

clayrosenthal: I think `~=` would be pretty reasonable

kingb: Added <2

kingb:
> Why do we need the nvidia driver `550.54.14` -> `555.58.02` bump? Is it jetson specific?

Driver 550 had a huge bug... wanted to avoid

PR #73, opened 2026-03-10
Added dual policy mode to toggle default locomotion policy

carlosferrazza:
This PR adds as a default inference config the option to switch to a robust locomotion policy by pressing X on the joystick (or x on keyboard), which is convenient for the robot to recover while testing experimental policies. The implementation is even more general than that, as the way it works is by introducing a dual mode setup that allows switching between any two policies at runtime.

### DIFF
--- a/tests/e2e/test_run_policy.py
+++ b/tests/e2e/test_run_policy.py
@@ -71,7 +71,8 @@ def assert_run_policy_with_hsinference(config_name: str, model_path: str, timeou
             f"pip install -e {REPO_ROOT}/src/holosoma_inference[unitree,booster] && "
             f"python {REPO_ROOT}/src/holosoma_inference/holosoma_inference/run_policy.py "
             f"inference:{config_name} "
-            f"--task.model-path={model_path}",
+            f"--task.model-path={model_path} "
+            f"--secondary none",
         ],
         stdin=subprocess.DEVNULL,
         stdout=subprocess.PIPE,


src/holosoma_inference/holosoma_inference/config/config_values/task.py
```diff
@@ -2,8 +2,12 @@
 
 from __future__ import annotations
 
+from pathlib import Path
+
 from holosoma_inference.config.config_types.task import TaskConfig
 
+_MODELS_DIR = Path(__file__).parent.parent.parent / "models"
```
clayrosenthal, line 9: could you add a comment clarifying where this should point to? Seems like `src/holosoma_inference/holosoma_inference/models`

src/holosoma_inference/holosoma_inference/run_policy.py
```diff
@@ -117,11 +131,89 @@ def run_policy(config: InferenceConfig):
...
+            expect_secondary_value = False
+        else:
+            primary.append(arg)
+            expect_secondary_value = False
+    return primary, secondary
+
+
 def main(annotated_config=None):
     """Main entry point. Extensions can pass their own AnnotatedInferenceConfig."""
+    import argparse
```
clayrosenthal, line 161: why not import at the top of the file?

src/holosoma_inference/holosoma_inference/run_policy.py
```diff
@@ -117,11 +131,89 @@ def run_policy(config: InferenceConfig):
         restore_terminal_settings()
 
 
+def _split_secondary_args(argv: list[str]) -> tuple[list[str], list[str]]:
```
tomasz-lewicki, line 134: Doesn't `tyro` handle this out-of-the box once you've defined `TaskConfig.secondary`? Maybe I'm missing something, but I think you can do:

```bash
run_policy.py --task.secondary.task.model-path weights.onnx
```

We can also just have a secondary optional `InferenceConfig`.
```bash
run_policy.py  --task.model-path weights_dance.onnx  --fallback_task.model-path weights_loco.onnx
```

src/holosoma_inference/holosoma_inference/policies/dual_mode.py
```diff
@@ -0,0 +1,166 @@
...
+        self._orig_joy = {
+            id(self.primary): self.primary.handle_joystick_button,
+            id(self.secondary): self.secondary.handle_joystick_button,
+        }
+        self._orig_kb = {
+            id(self.primary): self.primary.handle_keyboard_button,
+            id(self.secondary): self.secondary.handle_keyboard_button,
+        }
+
+        def patched_joy(cur_key):
```
tomasz-lewicki, line 92: Hopefully we refactor joystick/input handling soon so doing simple things doesn't have to be so convoluted.

PR #101, opened 2026-04-20
[eval callback]Refactor eval callbacks into grid-based multi-condition sweep architecture

Juyue:
Replace single-env push/payload callbacks with grid evaluation system that sweeps across multiple conditions in parallel. Velocity is always the base axis; either payload or push can be cross-producted with it (but not both, since they share the IsaacSim external force API).

Key changes:
- GridConditionManager: unified add_axis() API for Cartesian product of sweep axes, shared across all callbacks
- EvalRecordingCallback: single recorder with ensure_finalized() pattern for order-independent deferred setup
- GridEvalVelocityCallback: independent velocity sweep (lin_x/y, ang_yaw)
- GridEvalPayloadCallback: body-group x mass sweep with per-substep force injection via monkey-patched physics step
- GridEvalPushCallback: body x direction x gait_phase x force sweep with GaitAnalyser for phase-triggered pushes
- GaitAnalyser: standalone gait cycle detection from foot heights
- sim_utils: resolve_body(), apply_body_force_world(), clear_body_force()
- Config validation via @model_validator (body/label lengths, directions, gait phases, mutual exclusion of payload+push)

Minor:
- Update src/holosoma/holosoma/config_values/loco/g1/reward.py to reduce visual noise.


src/holosoma/holosoma/agents/callbacks/base_callback.py
```diff
@@ -8,6 +12,19 @@ def __init__(self, config, training_loop):
...
+        return self.training_loop._unwrap_env()
+
+    def _require_recording_cb(self) -> Any:
+        """Find and return the EvalRecordingCallback, or raise."""
+        from holosoma.agents.callbacks.recording import EvalRecordingCallback
+
+        for cb in self.training_loop.eval_callbacks:
+            if isinstance(cb, EvalRecordingCallback):
+                return cb
+        raise RuntimeError(f"{type(self).__name__} requires EvalRecordingCallback. Set --recording.config.enabled=True")
```
samuelgundry: Will there ever be a second eval CB? defensively, it can help to check here and error if so.

src/holosoma/holosoma/agents/callbacks/gait_analysis.py
```diff
@@ -0,0 +1,172 @@
...
+        if cycles < self._min_gait_cycles:
+            logger.warning(
+                f"GaitAnalyser: forced finalization with only "
+                f"{cycles} cycles (need {self._min_gait_cycles}). "
+                f"Phase detection may be unreliable."
+            )
+
+        self._left_foot_z_range = (np.min(left_z, axis=0), np.max(left_z, axis=0))
+        self._right_foot_z_range = (np.min(right_z, axis=0), np.max(right_z, axis=0))
+        self._done = True
```
samuelgundry: perhaps set done at the end of the function, just in case of confusion w.r.t debugging

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
```diff
@@ -0,0 +1,195 @@
...
+        self._axes: list[dict[str, Any]] = []
+        self.conditions: list[dict[str, Any]] = []
+        self.num_conditions: int = 0
+        self.warmup_steps: int = 0
+        self._finalized: bool = False
+
+    def add_axis(
+        self,
+        name: str | list[str],
+        values: list,
```
samuelgundry: what happens for duplicate axis names?  "add" vs. "register".  Good to document the behavior/expectations. You append below without checking but idk (yet) if you expect a single name. 

edit: i see below you de-dupe. Good to document then, especially for callers -- right now, there's no way to know a duplication happened and the match is by key only? not values

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
```diff
@@ -0,0 +1,195 @@
...
+                for keys, vals in zip(key_lists, combo):
+                    cond.update(dict(zip(keys, vals)))
+                self.conditions.append(cond)
+
+        # --- Deduplicate (preserving order) ---
+        seen: set[tuple] = set()
+        unique: list[dict[str, Any]] = []
+        for cond in self.conditions:
+            key = tuple(sorted(cond.items()))
+            if key not in seen:
```
samuelgundry: I suspect at the least you'll want to log duplicates, but see above re: comparing keys vs. values, and what callers can/can't do if there are duplicates with different values

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
```diff
@@ -0,0 +1,195 @@
...
+
+        ``grid_conditions`` uses hierarchical dicts grouped by the ``group``
+        parameter passed to ``add_axis()``.  Ungrouped keys stay at
+        the top level.
+
+        Example condition::
+
+            {'velocity': {'lin_vel_x': 0.5, 'lin_vel_y': 0.0, 'ang_vel_yaw': 0.0},
+             'push': {'body_label': 'torso', 'direction': 'forward', 'force_n': 150.0}}
+        """
```
samuelgundry, line 171: I don't know our current position towards unit testing, but this kind of logic is a good reason to add them, especially for future regressions.

samuelgundry (commented):
Part-way through, up to src/holosoma/holosoma/agents/callbacks/grid_eval_payload.py

src/holosoma/holosoma/agents/callbacks/base_callback.py
Juyue: 1. Theoretically, no second EvalRecordingCallback shows up.
2. Added check to be denfensive.

src/holosoma/holosoma/agents/callbacks/gait_analysis.py
Juyue: Thanks for suggestions. Done

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
Juyue: 1. Register seems more appropriate. 
2. Add check for duplicated axes name.
3. Move the dedup code to where it actually belong (src/holosoma/holosoma/agents/callbacks/grid_eval_velocity.py)

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
Juyue: Good call!  

Now, I move the dedup code from `grid_conditions` to where it actually belong (`src/holosoma/holosoma/agents/callbacks/grid_eval_velocity.py`). 

Grid_conditions now just does cartesian product without .

src/holosoma/holosoma/agents/callbacks/grid_conditions.py
Juyue: Yes, this part is a bit messy.
1. To simplify the code, I removed some unused/convenience meta data.
2. Instead of having a unit test, I have a doc-string to demonstrate the intended behavior.

PR #105, opened 2026-04-28
feat: add uv-based setup + activation scripts for holosoma_inference

kingb:
## Summary

Add `scripts/setup_inference_via_uv.sh` + `scripts/source_inference_uv_setup.sh`, mirroring the existing `setup_mujoco_via_uv.sh` pattern but for `holosoma_inference`. Creates an isolated uv-managed venv at `.venv/hsinference` instead of the system Python / conda.

## Why

Two concrete reasons:

1. **Faster installs** — `uv` resolves and installs wheels ~10-100× faster than `pip`, especially on slow networks.
2. **System-Python friendly** — JetPack 7.1 / Ubuntu 24.04 ships Python 3.12 with PEP 668 protections; users need a venv, and conda is overkill for pure-pip workflows.

## Changes

- `scripts/setup_inference_via_uv.sh`: installs `holosoma_inference` (editable), robot SDK extras on Linux, and Pinocchio (`pin>=3.8.0`) for WBT policy support. Auto-detects Python version (3.10 on Ubuntu 22.04, 3.12 on 24.04) or accepts `--python` override.
- `scripts/source_inference_uv_setup.sh`: activation script; sources ROS 2 if `/opt/ros/{jazzy,humble}` is present, then activates the venv.

CLI flags match the mujoco uv script for consistency: `--python VERSION`, `--no-robot-sdks`, `--reinstall`, `--help`.

## Test plan

- [x] Smoke-tested on Ubuntu 22.04 x86_64 (Python 3.10) — `holosoma_inference` imports, pinocchio 3.9.0 installs.
- [x] Smoke-tested on Ubuntu 24.04 aarch64 (Python 3.12, on Jetson AGX Thor) — same validations.
- [x] `source scripts/source_inference_uv_setup.sh` activates the venv.

## Known caveat

Activating the venv *after* sourcing ROS 2 Humble can cause `import pinocchio` to segfault due to Boost/numpy ABI conflicts between ROS Humble's system libs and pinocchio's bundled deps. Not a regression (same issue exists for `setup_mujoco_via_uv.sh`). Mitigation: activate the venv *first* if pinocchio is needed, or use a ROS 2 Jazzy environment where the system numpy matches.

### DIFF
--- a/scripts/setup_inference_via_uv.sh
+++ b/scripts/setup_inference_via_uv.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+# Exit on error, and print commands
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$(dirname "$SCRIPT_DIR")
+
+# Venv configuration
+VENV_DIR=$ROOT_DIR/.venv/hsinference
+
+# Parse command-line arguments
+INSTALL_ROBOT_SDKS=true
+PYTHON_VERSION=""
+REINSTALL=false
+
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --no-robot-sdks)
+      INSTALL_ROBOT_SDKS=false
+      echo "Robot SDK installation disabled (unitree, booster)"
+      shift
+      ;;
+    --python)
+      PYTHON_VERSION="$2"
+      shift 2
+      ;;
+    --reinstall)
+      REINSTALL=true
+      echo "Reinstall requested — existing environment will be removed"
+      shift
+      ;;
+    --help|-h)
+      echo "Usage: $0 [--no-robot-sdks] [--python VERSION] [--reinstall]"
+      echo ""
+      echo "Options:"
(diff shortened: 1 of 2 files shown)


scripts/setup_inference_via_uv.sh
```diff
@@ -0,0 +1,169 @@
+#!/bin/bash
+# Exit on error, and print commands
+set -e
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$(dirname "$SCRIPT_DIR")
+
+# Venv configuration
+VENV_DIR=$ROOT_DIR/.venv/hsinference
```
clayrosenthal: perhaps this could be like:
```
VENV_DIR="${HS_INFER_VENV:-$ROOT_DIR/.venv/hsinference}"
```
So we could change exactly where we want to install the venv?

kingb: Good call — done in 2f2d52a. Added `HS_INFER_VENV` override to both `setup_inference_via_uv.sh` and `source_inference_uv_setup.sh` so they stay in sync.

tomasz-lewicki (approved):
LGTM

PR #107, opened 2026-05-05
Cache weights for inference

tomasz-lewicki:
## Desc

This PR adds caching W&B checkpoint downloads under `~/.cache/holosoma_inference/<run_id>/<filename>` so `run_policy.py` reuses previously downloaded ONNX models instead of re-fetching them from W&B on every invocation.


Juyue (approved):
Thanks. (In existing description, it is still saying "~/.hs_weights/<run_id>/<filename>", update before merge.

PR #110, opened 2026-05-14
Add generic joystick support as input 

tomasz-lewicki:
This PR adds joystick controller as an input for `holosoma_inference`. Tested with Xbox controller.

### DIFF
--- a/src/holosoma_inference/holosoma_inference/config/config_types/task.py
+++ b/src/holosoma_inference/holosoma_inference/config/config_types/task.py
@@ -67,9 +67,20 @@ class TaskConfig:
     """
 
     use_joystick: bool = False
-    """Shortcut: set both velocity_input and state_input to "joystick".
+    """Shortcut: set both velocity_input and state_input to "interface".
+
+    Reads from the SDK's wireless controller (the dongle/controller shipped
+    with Unitree G1, Booster T1, etc.). For host-side USB gamepads
+    (Xbox/Logitech via /dev/input/event*), use ``use_usb_joystick`` instead.
 
     Cannot be combined with explicit input settings.
+    """
+
+    use_usb_joystick: bool = False
+    """Shortcut: set both velocity_input and state_input to "joystick".
+
+    Reads a USB gamepad on the host via evdev (``/dev/input/event*``).
+    Linux-only. Cannot be combined with explicit input settings.
     """
 
     joystick_type: str = "xbox"
@@ -113,17 +124,30 @@ class TaskConfig:
     """Debug overrides for quick testing."""
 
     def __post_init__(self):
-        """Resolve use_keyboard/use_joystick shortcuts into velocity_input/state_input."""
-        if self.use_keyboard and self.use_joystick:
+        """Resolve use_keyboard/use_joystick/use_usb_joystick shortcuts into velocity_input/state_input."""
+        active_shortcuts = [
+            name
+            for name, enabled in (
+                ("keyboard", self.use_keyboard),
+                ("joystick", self.use_joystick),
+                ("usb-joystick", self.use_usb_joystick),
+            )
+            if enabled
+        ]
+        if len(active_shortcuts) > 1:
+            joined = ", ".join(f"--task.use-{n}" for n in active_shortcuts)
             raise ValueError(
-                "Cannot combine --task.use-keyboard with --task.use-joystick. "
+                f"Cannot combine multiple input shortcuts ({joined}). "
                 "Use one shortcut or set --task.velocity-input and --task.state-input individually."
             )
 
         shortcut: InputSource | None = None
         flag_name: str | None = None
-        if self.use_joystick:
+        if self.use_usb_joystick:
             shortcut = "joystick"
+            flag_name = "usb-joystick"
+        elif self.use_joystick:
+            shortcut = "interface"
             flag_name = "joystick"
         elif self.use_keyboard:
             shortcut = "keyboard"
--- a/src/holosoma_inference/holosoma_inference/config/config_types/task.py
+++ b/src/holosoma_inference/holosoma_inference/config/config_types/task.py
@@ -123,7 +123,7 @@ class TaskConfig:
         shortcut: InputSource | None = None
         flag_name: str | None = None
         if self.use_joystick:
-            shortcut = "interface"
+            shortcut = "joystick"
             flag_name = "joystick"
         elif self.use_keyboard:
             shortcut = "keyboard"
--- a/src/holosoma_inference/holosoma_inference/inputs/__init__.py
+++ b/src/holosoma_inference/holosoma_inference/inputs/__init__.py
@@ -17,8 +17,13 @@ def create_input(policy: BasePolicy, source: InputSource, role: str) -> VelCmdPr
     if not policy.use_joystick and source in ("interface", "joystick"):
         source = "keyboard"
 
-    if source in ("interface", "joystick"):
+    if source == "interface":
         return InterfaceInput(policy.interface)
+
+    if source == "joystick":
+        from holosoma_inference.inputs.impl.usb_joystick import UsbJoystickInput
+
+        return UsbJoystickInput(device_index=policy.config.task.joystick_device)
 
     if source == "keyboard":
         vel_keys = KEYBOARD_VELOCITY_LOCOMOTION if role == "velocity" else None
--- a/src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
+++ b/src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
@@ -1,0 +1,270 @@
+"""USB joystick input provider (evdev-based).
+
+Reads a USB gamepad (Xbox / Logitech / similar) directly from
+``/dev/input/event*`` via ``python-evdev``.  Bypasses the SDK
+:class:`InterfaceInput` path so that SDKs without a built-in wireless
+controller (e.g. BXI ELF3) can still drive policies from a host-side
+controller.
+
+Implements both :class:`VelCmdProvider` and :class:`StateCommandProvider`
+in a single class — the policy factory assigns the same instance to both
+slots when ``velocity_input == state_input == "joystick"``.
+"""
+
+from __future__ import annotations
+
+import threading
+
+import evdev
+
+from holosoma_inference.inputs.api.base import InputProvider
+from holosoma_inference.inputs.api.commands import StateCommand, VelCmd
+from holosoma_inference.inputs.impl.joystick import JOYSTICK_COMMANDS
+
+STICK_DEADZONE = 0.1
+TRIGGER_THRESHOLD = 128  # 0–255 typical for analog triggers; >threshold counts as pressed.
+
+# Match interface_wrapper.py's _default_wc_key_map bit layout so policies that
+# inspect raw key codes still see consistent values.
+_BIT_R1 = 1
+_BIT_L1 = 2
+_BIT_START = 4
+_BIT_SELECT = 8
+_BIT_R2 = 16
+_BIT_L2 = 32
+_BIT_A = 256
+_BIT_B = 512
+_BIT_X = 1024
+_BIT_Y = 2048
+_BIT_UP = 4096
+_BIT_RIGHT = 8192
+_BIT_DOWN = 16384
+_BIT_LEFT = 32768
+
+_BUTTON_BIT = {
+    evdev.ecodes.BTN_A: _BIT_A,
+    evdev.ecodes.BTN_B: _BIT_B,
+    evdev.ecodes.BTN_X: _BIT_X,
+    evdev.ecodes.BTN_Y: _BIT_Y,
+    evdev.ecodes.BTN_TL: _BIT_L1,
+    evdev.ecodes.BTN_TR: _BIT_R1,
+    evdev.ecodes.BTN_TL2: _BIT_L2,
+    evdev.ecodes.BTN_TR2: _BIT_R2,
+    evdev.ecodes.BTN_START: _BIT_START,
+    evdev.ecodes.BTN_SELECT: _BIT_SELECT,
+}
+
+# Subset of the Unitree wireless-controller bitmask map sufficient for the
+# combinations referenced in JOYSTICK_COMMANDS.  Kept inline so this module
+# does not depend on the SDK.
+_KEY_LABEL = {
+    _BIT_R1: "R1",
+    _BIT_L1: "L1",
+    _BIT_L1 | _BIT_R1: "L1+R1",
+    _BIT_START: "start",
+    _BIT_SELECT: "select",
+    _BIT_R2: "R2",
+    _BIT_L2: "L2",
+    _BIT_A: "A",
+    _BIT_SELECT | _BIT_A: "select+A",
+    _BIT_B: "B",
+    _BIT_SELECT | _BIT_B: "select+B",
+    _BIT_X: "X",
+    _BIT_SELECT | _BIT_X: "select+X",
+    _BIT_Y: "Y",
+    _BIT_SELECT | _BIT_Y: "select+Y",
+    _BIT_UP: "up",
+    _BIT_DOWN: "down",
+    _BIT_LEFT: "left",
+    _BIT_RIGHT: "right",
+}
+
+
+def _list_gamepads() -> list[evdev.InputDevice]:
+    """Return evdev devices that look like gamepads (have sticks + buttons)."""
+    candidates: list[evdev.InputDevice] = []
+    for path in evdev.list_devices():
+        try:
+            dev = evdev.InputDevice(path)
+        except (PermissionError, OSError):
+            continue
+        caps = dev.capabilities()
+        if evdev.ecodes.EV_ABS not in caps or evdev.ecodes.EV_KEY not in caps:
+            dev.close()
+            continue
+        abs_codes = {code for code, _ in caps[evdev.ecodes.EV_ABS]}
+        if {evdev.ecodes.ABS_X, evdev.ecodes.ABS_Y, evdev.ecodes.ABS_RX} <= abs_codes:
+            candidates.append(dev)
+        else:
+            dev.close()
+    return candidates
+
+
+class UsbJoystickInput(InputProvider):
+    """Reads stick + button state from a USB gamepad via evdev.
+
+    A daemon thread continuously consumes events and updates internal state.
+    :meth:`poll_velocity` returns the latest sticks; :meth:`poll_commands`
+    edge-detects button transitions and emits :class:`StateCommand` values
+    using :data:`JOYSTICK_COMMANDS`.
+    """
+
+    def __init__(self, device_index: int = 0):
+        self._mapping = dict(JOYSTICK_COMMANDS)
+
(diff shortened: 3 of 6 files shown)


src/holosoma_inference/holosoma_inference/config/config_types/task.py
```diff
@@ -123,7 +123,7 @@ def __post_init__(self):
         shortcut: InputSource | None = None
         flag_name: str | None = None
         if self.use_joystick:
-            shortcut = "interface"
+            shortcut = "joystick"
```
kingb, line 147: **Behavior change** — see the top-level review comment. This single character change (`"interface"` → `"joystick"`) silently flips the meaning of `--task.use-joystick` for every existing G1/T1 user from "SDK wireless controller" to "USB gamepad on the host." Please confirm intent and either:

- add a fallback so `"joystick"` reverts to `"interface"` when no USB gamepad is present, or
- preserve `--task.use-joystick → "interface"` and add a separate `--task.use-usb-joystick` shortcut for the new behavior.

Whichever direction we pick, please update the `use_joystick` docstring just above to spell it out.

src/holosoma_inference/holosoma_inference/inputs/__init__.py
```diff
@@ -17,9 +17,14 @@ def create_input(policy: BasePolicy, source: InputSource, role: str) -> VelCmdPr
...
         source = "keyboard"
 
-    if source in ("interface", "joystick"):
+    if source == "interface":
         return InterfaceInput(policy.interface)
 
+    if source == "joystick":
+        from holosoma_inference.inputs.impl.usb_joystick import UsbJoystickInput
+
+        return UsbJoystickInput(device_index=policy.config.task.joystick_device)
```
kingb, line 26: Lazy-importing `UsbJoystickInput` here is the right call — it keeps macOS users (where `evdev` doesn't build) able to import this module as long as they don't request `"joystick"`. ✓

One thing to watch: the test in `inputs/tests/test_providers.py` does `from holosoma_inference.inputs.impl import usb_joystick` at the top of the test method, which will fail to collect on macOS regardless. Suggest `pytest.importorskip("evdev")` there.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+Implements both :class:`VelCmdProvider` and :class:`StateCommandProvider`
+in a single class — the policy factory assigns the same instance to both
+slots when ``velocity_input == state_input == "joystick"``.
+"""
+
+from __future__ import annotations
+
+import threading
+
+import evdev
```
kingb, line 17: **macOS test-collection failure:** importing this module on macOS crashes at this line because `evdev` only builds on Linux (it links against `<linux/input.h>`). The lazy import in `inputs/__init__.py` keeps production code safe, but `tests/test_providers.py::test_joystick_maps_to_usb_joystick` does `from holosoma_inference.inputs.impl import usb_joystick` unconditionally, which will fail collection on darwin.

Two options:
1. **Easier:** at the top of the test, `pytest.importorskip("evdev")`.
2. **More invasive:** defer `import evdev` to inside the function bodies that use it, so this module can be imported on darwin (just not instantiated).

Option 1 is sufficient.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            dev = evdev.InputDevice(path)
+        except (PermissionError, OSError):
+            continue
+        caps = dev.capabilities()
+        if evdev.ecodes.EV_ABS not in caps or evdev.ecodes.EV_KEY not in caps:
+            dev.close()
+            continue
+        abs_codes = {code for code, _ in caps[evdev.ecodes.EV_ABS]}
+        if {evdev.ecodes.ABS_X, evdev.ecodes.ABS_Y, evdev.ecodes.ABS_RX} <= abs_codes:
```
kingb, line 96: Requiring `ABS_RX` excludes single-stick / flight-style controllers but is fine for Xbox/Logitech-class gamepads. Worth a docstring note that this is gamepad-only by design — otherwise the next debug session for "my controller wasn't detected" goes through this whole function.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+            candidates.append(dev)
+        else:
+            dev.close()
+    return candidates
```
kingb, line 100: **Device identity is fragile.** `evdev.list_devices()` returns `/dev/input/event*` in alphabetical order, which is *not* stable across reboots/replugs. Today's `joystick_device=0` could be tomorrow's `=1`.

For a single-controller setup it's fine; for multi-controller or robust-replug setups it's a footgun.

Suggestion (potentially future-CR): add an optional `joystick_device_id: str = ""` config field with this match order:

1. `dev.uniq` exact match (when populated, this is the closest evdev equivalent to a MAC — Bluetooth gamepads almost always have it; some wired USB gamepads do too if the vendor bothered with serials).
2. Substring match against the `/dev/input/by-id/usb-VENDOR_PRODUCT_SERIAL-event-joystick` symlink target.
3. Substring match against `dev.name` (warn on >1 match).
4. Fall back to `joystick_device` index when `joystick_device_id` is empty (today's behavior).

`dev.uniq` is the strongest primitive when populated — worth exploring even if we don't ship the full chain in this PR.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+        gamepads = _list_gamepads()
+        if not gamepads:
+            raise RuntimeError(
+                "No USB joystick found via evdev. Is the controller plugged in "
+                "and is /dev/input mounted into this container?"
+            )
+        if device_index >= len(gamepads):
+            for d in gamepads:
+                d.close()
+            raise RuntimeError(f"joystick_device={device_index} but only {len(gamepads)} gamepad(s) detected")
```
kingb, line 127: Negative `device_index` (e.g. `-1`) passes this check and silently selects the last gamepad due to Python list semantics — probably not what callers expect. Suggest:

```python
if device_index < 0 or device_index >= len(gamepads):
    ...
```

kingb (commented):
Thanks for putting this together — the host-side evdev path is well-isolated and the threading model looks right. **Net: approve with one must-fix and a handful of smaller items.**

Single biggest concern, calling out at the top because it isn't obvious from the diff:

## ⚠️ Behavior change in `--task.use-joystick`

Before this PR, `--task.use-joystick` resolved to `velocity_input = state_input = "interface"` — i.e. the **SDK wireless controller** (the dongle/controller shipped with Unitree G1, Booster T1, etc.). After this PR it resolves to `"joystick"` → `UsbJoystickInput` reading `/dev/input/event*`.

For a user running on a real G1 with the Unitree controller and no USB gamepad, the same command they ran yesterday will now error out with `RuntimeError: No USB joystick found via evdev`. This is a silent semantic flip of a user-facing flag, and I want to make sure it's intentional before merge.

Questions / suggestions:
1. **Is this intentional?** If the goal is "on Thor, default to a host-side gamepad," that's a reasonable design — but it should be a deliberate decision, not a side effect.
2. **The old behavior is still reachable** via `--task.velocity-input interface --task.state-input interface`, but nothing surfaces that to the user.
3. **Suggestion** — keep the new `"joystick"` source, but pick one of:
   - Have `--task.use-joystick` first try `UsbJoystickInput`, and if no USB gamepad is present, fall back to `"interface"`. (Mirrors the existing darwin → keyboard fallback in `_init_joystick_handler`.)
   - Add an explicit `--task.use-sdk-controller` shortcut for the old behavior, keep `--task.use-joystick` meaning what it now means, and add a release note.
   - Or rename the new shortcut to `--task.use-usb-joystick` and leave `--task.use-joystick` pointing at `"interface"`.
   
   Whatever we pick, the PR description and the `TaskConfig.use_joystick` docstring should call out the change.

---

### Must-fix

- Test will fail to collect on macOS — `usb_joystick.py` does `import evdev` at module top level, and `evdev` only builds on Linux. The new `test_joystick_maps_to_usb_joystick` imports the module unconditionally. Wrap with `pytest.importorskip("evdev")` or defer the `import evdev` inside `usb_joystick`. (Inline.)
- Negative `device_index` is silently accepted and indexes from the end. (Inline.)

### Should-fix

- Device identity is fragile across replugs — `_list_gamepads()` orders by `evdev.list_devices()` (alphabetical over `/dev/input/event*`), so today's index 0 may be tomorrow's index 1. `dev.uniq` (potentially populated with a serial / BT MAC) is the right primitive when available; see inline.
- `start()` is a no-op because the thread is started in `__init__` — breaks the protocol's lifecycle contract.
- `_KEY_LABEL` doesn't include `F1` (keyboard fallback, fine) or lowercase `x` (looks like a typo in `JOYSTICK_COMMANDS`). Pre-existing in `joystick.py`, worth a follow-up.

### Out-of-scope / future-CR candidates

- `STICK_DEADZONE = 0.1` now duplicated here + `interface.py`. Pull into `inputs/impl/joystick.py` next to `JOYSTICK_COMMANDS`.
- `joystick_type: str = "xbox"` is unused — either wire it through (per-controller layout) or drop it.
- The stick-suppression-while-button-held behavior is inherited from `InterfaceInput`; the rationale there was that the SDK wire format couples sticks + keys in one frame, which doesn't apply to evdev. Current behavior freezes commanded velocity mid-chord — benign but not necessarily what the user wants. Worth re-examining together with `InterfaceInput`.

### Nit

- PR description says "Tested with Xbox controller" — please add the model (Series X? 360? wired vs wireless via dongle?). Different generations expose different evdev codes especially on triggers.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+        self._rx = 0.0  # right stick X, normalized [-1, 1], left=+ after sign flip
+        self._key_bits = 0  # OR'd _BIT_* of currently-held buttons
+        self._dpad_x = 0  # -1 / 0 / 1 from ABS_HAT0X
+        self._dpad_y = 0  # -1 / 0 / 1 from ABS_HAT0Y
+
+        self._last_label: str = ""
+
+        self._stop = threading.Event()
+        self._thread = threading.Thread(target=self._run, daemon=True, name="usb_joystick")
+        self._thread.start()
```
kingb, line 153: The thread is started here in `__init__`, which makes `start()` (below) a no-op. That breaks the `InputProvider` lifecycle contract — `BasePolicy._create_input_providers()` calls `self._velocity_input.start()` *after* construction expecting that's when the input goes live.

Suggest moving thread creation into `start()` and having `__init__` only enumerate + select the device. Matches `KeyboardInput` semantics and means construction is cheap/safe (e.g. for tests, fixtures, etc.).

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def start(self) -> None:
+        pass  # Thread already running from __init__.
+
+    def close(self) -> None:
+        self._stop.set()
+        try:
+            self._device.close()
+        except OSError as e:
+            logger.debug(f"evdev close raised {type(e).__name__}: {e}")
```
kingb, line 165: `close()` isn't currently called by anyone — `InputProvider` protocol has `start()` but no `close()`, and policies don't tear down their input providers explicitly. The daemon thread will be killed at process exit, which is fine for a long-running policy.

Worth a comment that this is intentional, or — better — register an `atexit` handler so the device fd doesn't linger if multiple `UsbJoystickInput`s are constructed (tests, accidental double-init, etc.).

Related: `_list_gamepads()` opens every `/dev/input/event*` and closes all but the chosen one. On a host with bluetoothd / accessibility devices that's 20+ opens — a `try/finally` around the close calls would prevent fd leaks if a `RuntimeError` fires mid-scan.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+    # -- VelCmdProvider protocol -----------------------------------------
+
+    def poll_velocity(self) -> VelCmd | None:
+        with self._lock:
+            keys = self._effective_key_bits_locked()
+            lx, ly, rx = self._lx, self._ly, self._rx
+
+        # Match InterfaceInput: suppress sticks while any button is held.
+        if keys != 0:
+            return None
```
kingb, line 176: Re: stick suppression while a button is held — flagging for future discussion, not this PR.

This mirrors `InterfaceInput.poll_velocity`, which suppresses sticks because the SDK packs sticks + keys into a single wireless-controller frame and the SDK's own button-handling steals the frame. That rationale doesn't apply to evdev (sticks and buttons are independent event streams), so we're inheriting a UX choice here, not a hardware constraint.

Net effect: holding any button freezes the commanded velocity at whatever it last was. Benign (the robot keeps doing what it was doing), but not necessarily desirable for chord-while-moving flows. Worth re-examining as a follow-up — and if we change it, change `InterfaceInput` at the same time.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+            keys = self._effective_key_bits_locked()
+        label = _KEY_LABEL.get(keys, "")
+
+        commands: list[StateCommand] = []
+        if label and label != self._last_label:
+            cmd = self._mapping.get(label)
+            if cmd is not None:
+                commands.append(cmd)
+        self._last_label = label
+        return commands
```
kingb, line 199: Edge detection on `_last_label` can swallow rapid same-chord presses if both the press and release land between two `poll_commands()` cycles (~policy `rl_rate`, usually 50Hz). `InterfaceInput` has the same model and same theoretical bug, so not a regression — just noting that command repeat-rate is bounded by policy rate, not physical button rate.

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+        except OSError:
+            return  # Device unplugged.
+
+    def _handle_abs(self, code: int, value: int) -> None:
+        if code in self._abs_info:
+            info = self._abs_info[code]
+            span = info.max - info.min
+            if span <= 0:
+                return
+            normalized = (value - info.min) / span * 2.0 - 1.0  # → [-1, 1]
```
kingb, line 221: Operator precedence: `(value - info.min) / span * 2.0 - 1.0` evaluates as `((value - info.min) / span) * 2.0 - 1.0` — correct, but adding outer parens reads more clearly:

```python
normalized = ((value - info.min) / span) * 2.0 - 1.0
```

src/holosoma_inference/holosoma_inference/inputs/impl/usb_joystick.py
```diff
@@ -0,0 +1,270 @@
...
+        elif code == evdev.ecodes.ABS_HAT0X:
+            with self._lock:
+                self._dpad_x = int(value)
+        elif code == evdev.ecodes.ABS_HAT0Y:
+            with self._lock:
+                self._dpad_y = int(value)
+        elif code == evdev.ecodes.ABS_Z:  # analog L2
+            self._set_bit(_BIT_L2, value > TRIGGER_THRESHOLD)
+        elif code == evdev.ecodes.ABS_RZ:  # analog R2
+            self._set_bit(_BIT_R2, value > TRIGGER_THRESHOLD)
```
kingb, line 238: `TRIGGER_THRESHOLD = 128` assumes an unsigned 0–255 range. Some controllers (especially older Xbox 360 wired pads) report `ABS_Z`/`ABS_RZ` as signed -32768..32767, in which case `value > 128` is true for the entire "trigger pressed past quarter" range including the resting position on some pads.

Safer: read `info.min` / `info.max` from `dev.capabilities()[EV_ABS]` and threshold at `min + (max - min) * 0.5`. Out of scope for this PR if Xbox-Series-X-only is the immediate target, but worth a TODO.

src/holosoma_inference/holosoma_inference/inputs/tests/test_providers.py
```diff
@@ -758,12 +758,16 @@ def test_interface_returns_interface_input(self):
         assert isinstance(result, InterfaceInput)
         assert result._mapping == JOYSTICK_COMMANDS
 
-    def test_joystick_maps_to_interface(self):
-        from holosoma_inference.inputs.impl.interface import InterfaceInput
+    def test_joystick_maps_to_usb_joystick(self, monkeypatch):
+        from holosoma_inference.inputs.impl import usb_joystick
+
+        sentinel = MagicMock()
+        monkeypatch.setattr(usb_joystick, "UsbJoystickInput", lambda **_: sentinel)
```
kingb, line 766: This unconditional import of `usb_joystick` will fail collection on macOS (where `evdev` doesn't build). Wrap with:

```python
import pytest
pytest.importorskip("evdev")
from holosoma_inference.inputs.impl import usb_joystick
```

or at the class level if other tests in this class touch the same module.

src/holosoma_inference/holosoma_inference/policies/base.py
```diff
@@ -149,10 +149,12 @@ def _init_communication_components(self):
...
             self.interface = self._shared_hardware_source.interface
             return
-        # Derive use_joystick for SDK: True if interface/joystick is used for either channel
+        # The SDK's own wireless-controller path is only needed when an
+        # "interface" channel is selected.  The "joystick" channel is read
+        # via host-side evdev (UsbJoystickInput) and does not touch the SDK.
         vel = self.config.task.velocity_input
         other = self.config.task.state_input
-        need_joystick = bool({"interface", "joystick"} & {vel, other})
+        need_joystick = "interface" in {vel, other}
```
kingb, line 157: This is the right semantic split — the SDK's wireless-controller path is only needed for `"interface"` channels now. ✓

Worth confirming on real hardware that a policy started with `--task.velocity-input interface` (no joystick channel) but `use_joystick=True` on the SDK still passes joystick-key events through correctly. The `need_joystick` boolean's meaning to `create_interface` is unchanged, but the *set* of code paths exercising it shrunk.

kingb (approved):
LGTM — thanks for addressing the must-fixes.

✅ `--task.use-joystick` semantics restored to `"interface"` (SDK wireless controller); new `--task.use-usb-joystick` shortcut for the evdev path. Docstring + tests both updated.
✅ Mutual-exclusion check generalized cleanly across all three shortcuts.
✅ Negative `device_index` rejected.
✅ macOS test collection unblocked via `pytest.importorskip("evdev")`.

Filed two follow-ups (not blocking):
- #113 — `start()` should be the point at which `UsbJoystickInput` goes live; today the thread is started in `__init__`, breaking the `InputProvider` lifecycle contract.
- #114 — stable device identity via `dev.uniq` matching, since `evdev.list_devices()` ordering isn't stable across replug.

Approving so this can land.

PR #111, opened 2026-05-14
Add multiprocessing based version of UnitreeInterface

tomasz-lewicki:
## Summary

Add `unitree_mp` as an opt-in `sdk_type` that runs the Unitree C++ binding in a spawned child process.
The proxy talks to the parent via queue-based RPC, which keeps `unitree_interface`'s bundled CycloneDDS isolated from any DDS already loaded by rclpy/ROS2 Jazzy, eliminating the `free(): invalid pointer` abort that occurs at DDS init when running policy code from a Jazzy-sourced shell.

The `unitree` default is unchanged. Two misc changes:
- `source_mujoco_py312_setup.sh` now prepends the bundled lib dir to `LD_LIBRARY_PATH`, which fixes the same crash for `run_sim.py` (whose simulator bridge is a separate code path that doesn't go through `holosoma.sdk` entry points).
- `holosoma_inference/sdk/interface_wrapper.py` deleted (no longer used, zero usage in repo)

## Testing
```bash
source /opt/ros/jazzy/setup.bash
source ~/projects/holosoma/scripts/source_mujoco_py312_setup.sh
python3 src/holosoma_inference/holosoma_inference/run_policy.py inference:g1-29dof-loco \
    --task.model-path src/holosoma_inference/holosoma_inference/models/loco/g1_29dof/fastsac_g1_29dof.onnx \
    --task.no-use-joystick --task.interface lo \
    --robot.sdk-type unitree_mp
```


src/holosoma_inference/holosoma_inference/sdk/unitree/unitree_interface_mp.py
```diff
@@ -0,0 +1,212 @@
...
+            except Exception as exc:
+                res_q.put(("err", exc))
+    finally:
+        # Drop the C++ binding ref before the worker function returns so its
+        # destructor runs while the DDS event loop is still alive. Then bypass
+        # Python's atexit chain — any lingering C++ teardown after this point
+        # tends to surface as a misleading `Process SpawnProcess-1:` stderr
+        # noise even though the parent already got a clean shutdown.
+        del robot
+        os._exit(0)
```
asetapen, line 129: Nit:  this also bypasses any cleanup the worker might want to do (logging flush, file close, future cleanup hooks added by users of `UnitreeInterface`).

src/holosoma_inference/holosoma_inference/sdk/unitree/unitree_interface_mp.py
```diff
@@ -0,0 +1,212 @@
...
+                    res_q.put(("ok", result))
+            except Exception as exc:
+                res_q.put(("err", exc))
+    finally:
+        # Drop the C++ binding ref before the worker function returns so its
+        # destructor runs while the DDS event loop is still alive. Then bypass
+        # Python's atexit chain — any lingering C++ teardown after this point
+        # tends to surface as a misleading `Process SpawnProcess-1:` stderr
+        # noise even though the parent already got a clean shutdown.
+        del robot
```
asetapen, line 128: Do we want to do an explicit `robot.close()` first?

src/holosoma_inference/holosoma_inference/sdk/unitree/unitree_interface_mp.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # ── lifecycle ──────────────────────────────────────────────────────
+
+    def close(self):
+        self._req_q.put(_STOP)
+        self._proc.join(timeout=5)
+        if self._proc.is_alive():
+            self._proc.kill()
+
+    def __del__(self):
```
asetapen, line 210: Ah, now I see the reasoning. At interpreter shutdown multiprocessing primitives could be already torn down, but I think this is fine if it works in practice.

src/holosoma_inference/holosoma_inference/sdk/unitree/unitree_interface_mp.py
```diff
@@ -0,0 +1,212 @@
...
+                ctypes.CDLL(str(lib_path), mode=ctypes.RTLD_GLOBAL)
+
+    from holosoma_inference.sdk.unitree.unitree_interface import UnitreeInterface
+
+    robot = UnitreeInterface(robot_config, domain_id, interface_str, use_joystick)
+
+    try:
+        while True:
+            msg = req_q.get()
+            if msg is _STOP:
```
asetapen, line 66: IF for some reason any other code path puts `None` on the queue, this will also stop. Just checking that's intended behavior. Otherwise you could do something like `_STOP = object()` so its a unique sentinel.

asetapen (approved):
A very useful addition to `holosoma`!

PR #116, opened 2026-05-27
Add Dampener into BaseInterface

asetapen:
## Summary

While testing learned policies on real hardware, there is often a need to dampen all outputs to keep the robot safe during development.  This PR implements an optional `Dampener` into `BaseInterface`. This dampener implements a concrete `send_low_command` that runs the shim, then forwards to a backend-only `_send_low_command_impl`. New backends inherit the dampener wiring for free; only the marshalling step remains backend-specific.

## Changelog

  - **`base_interface.py`** - `send_low_command`  calls `Dampener.apply` (when configured) then `_send_low_command_impl`. Adds `_build_dampener` and `_resolve_joint_limits` hooks so backends with an MJCF can supply per-joint hard limits to the dampener's `q_limit_scale` knob without leaking MJCF logic into the base. `load_joint_limits_from_mjcf` is exported for backends to opt in.
  - **`unitree_interface.py`** - renames `send_low_command` → `_send_low_command_impl`; overrides `_resolve_joint_limits` to call the MJCF loader; skips legacy `_kp_level`/`_kd_level` scaling when the dampener is configured (the dampener owns kp/kd scaling in that path) and keeps the legacy path intact for callers that mutate those fields.
  - **`booster_interface.py`** - same rename + same double-scale guard against the command sender's own `kp_level`/`kd_level`.
  - **`config_types/robot.py`** - adds optional `dampening: DampeningConfig | None = None` field. Default `None` means `send_low_command` is an identity transform - fully backwards compatible.
  - **Tests** - `tests/test_dampened_base.py` covers the template-method routing (dampener on/off paths, kp/kd not double-applied, slew + clip + blend interactions through the base). `tests/test_dampening.py` covers the dampener algebra. `tests/test_send_log.py` covers the JSONL ring logger.
  - **`send_log.py`** - separate but related: a structured per-frame send logger (env-gated, default off) for analyzing what actually went on the wire.
  - **README.md** - updates to document the multiprocessing unitree driver and the new dampener

  ## Backwards compatibility

  - Robots without `RobotConfig.dampening` see no behavior change: the base class skips the dampener path and calls `_send_low_command_impl` directly with the original args.
  - `kp_level` / `kd_level` remain on `UnitreeInterface` and `BoosterInterface.command_sender` and still scale gains in the non-dampened path. Legacy callers that mutate these fields are unaffected.

## Test plan

  - [x] `pytest src/holosoma_inference/holosoma_inference/sdk/base/tests/ src/holosoma_inference/holosoma_inference/sdk/tests/`
  - [x] Sim-to-sim G1 locomotion with `RobotConfig.dampening = None` - confirm identical behavior to main.
  - [x] Sim-to-sim G1 locomotion with `DampeningConfig(kp_level=0.5)` - confirm halved KP at the wire.
  - [x] Real-robot dry run with `HOLOSOMA_BLEND_ALPHA=0.0` - confirm output freezes at measured q.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ lint.ignore=[
 "src/holosoma/holosoma/train_agent.py" = ["PLC0415"]  # Disable since imports after SimApp are inherent to IsaacSim
 "src/holosoma/holosoma/utils/draw.py" = ["F401"]  # Disable since unused imports are required for the adapters
 "src/holosoma/holosoma/utils/eval_utils.py" = ["PLC0415"]  # Disable since imports after SimApp are inherent to IsaacSim
+"src/holosoma_inference/holosoma_inference/sdk/base/base_interface.py" = ["PLC0415"]  # Deferred imports for optional heavy deps (mujoco, holosoma_retargeting)
 "src/holosoma_inference/holosoma_inference/policies/base.py" = ["PLC0415"]  # Deferred imports for optional heavy deps (ROS2, joystick, keyboard providers)
 "src/holosoma_inference/holosoma_inference/policies/locomotion.py" = ["PLC0415"]  # Deferred provider imports
 "src/holosoma_inference/holosoma_inference/policies/wbt.py" = ["PLC0415"]  # Deferred provider imports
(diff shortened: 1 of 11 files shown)


tomasz-lewicki (changes requested):
Unless we genuinely use it across multiple applications, and multiple robots, we shouldn't add it to the default unitree/booster interface (even if it's disable by default).

How about we implement `DampenedUnitreeInterface` as a separate sdk and structure it either as core- or extension? This will significantly reduce the footprint of this change.

PR #126, opened 2026-06-18
fix(wbt): correct adaptive-timestep sampler

noahcao:
Multi-motion: derive motion_id from the global sampled frame (was random). Accumulate failure counts across per-step resets. Skip failure stats during eval.

*Issue #, if available:*
Previous adaptive sampling for multi-motion training has the index mismatch bug between global frame idx (on all motion clips concatenated) and the local frame idx inside a single clip.

*Description of changes:*
Use an explicit two-layer idx selection to avoid this issue and to make sure the fix applies to single-motion training as well.

### DIFF
--- a/src/holosoma/holosoma/managers/command/terms/wbt.py
+++ b/src/holosoma/holosoma/managers/command/terms/wbt.py
@@ -456,7 +456,11 @@ class AdaptiveTimestepsSampler:
             self.num_bins - 1,
         ).long()
         assert failed_bin.min() >= 0 and failed_bin.max() < self.num_bins, "Failed bin is out of range"
-        self.current_bin_failed_count[:] = torch.bincount(failed_bin, minlength=self.num_bins)
+        # Accumulate (not overwrite): reset() may be called more than once per env
+        # step — once for termination-driven resets and again for clip-ended resets
+        # in MotionCommand.step() — before update_bin_failed_count() folds + zeroes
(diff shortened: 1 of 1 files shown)


src/holosoma/holosoma/managers/command/terms/wbt.py
```diff
@@ -597,30 +614,52 @@ def reset(self, env_ids: torch.Tensor | None) -> None:
...
+            end_idx = self.motion.motion_end_idx[motion_ids]
+            self.time_steps[env_ids] = adaptive_global_idx.clamp(start_idx, end_idx - 1)
+        else:
+            # Uniform path (or eval): randomly assign each env to a motion, sample
+            # a phase within that motion's range.
+            self.motion_ids[env_ids] = torch.randint(0, num_motions, (n,), device=self.device)
+            start_idx = self.motion.motion_start_idx[self.motion_ids[env_ids]]
+            end_idx = self.motion.motion_end_idx[self.motion_ids[env_ids]]
+            motion_len = end_idx - start_idx
+            self.time_steps[env_ids] = start_idx + (phase * (motion_len - 1).float()).long()
```
carlosferrazza, line 662: There seems to be an issue here. In eval mode, phase may be None, and gets multiplied with a tensor

carlosferrazza (approved):
Just one comment below + failing tests to fix, else LGTM

PR #124, opened 2026-06-16
Run holosoma as a service

tomasz-lewicki:
# Summary / tl;dr 
This PR enables running holosoma as a service that takes an external input to the whole body tracker. 

1. Added `holosoma_inference_service` ros2 package (desc below)
2. Added APIs for high-level unitree clients (arm client and loco client specifically). `sdk/unitree_high_level`.
3. Added a generic [multiprocess_proxy.py](https://github.com/amazon-far/holosoma/pull/124/changes#diff-15a53f4c85da3016fc28246130794610403d25cd663cf2c46a944e269f136e1d) to deal with ros2 and unitree's cyclonedds overriding each other's symbols when loaded into the same process.

# Holosoma Inference Service

Service layer for deploying `holosoma` policies. The input/output API is ROS2 messages; a backend turns them into robot motion. The service is not required to run `holosoma_inference`, but can be helpful for integrating it into a larger system.

## Service API
Inputs (one of):

| Topic | Type | Description | Exp. Rate |
|---|---|---|---|
| `/holosoma/smplh_command` | `CmdSMPLH.msg` | SMPL-H 24-joint pose targets, retargeted to dense before the policy | 50Hz |
| `/holosoma/dense_tracking_command` | `CmdDense.msg` | Dense per-joint 29-DoF target (q/dq + root quat) consumed directly by the policy | 50Hz |
| `/holosoma/exoskeleton_command` | `CmdExoskeleton.msg` | Left/right arm joint targets (7+7) plus base twist for the split-body controller | 50Hz |
| `/holosoma/3pt_command` | `Cmd3pt.msg` | Head + wrist poses (not supported yet) | 50Hz |

Outputs:

| Topic | Type | Description | Rate |
|---|---|---|---|
| `/holosoma/holosoma_executed_cmd` | `JointState.msg` | Executed joint command, always full 29-DoF | Policy Rate (typ. 50Hz)  |
| `/holosoma/heartbeat` | `Heartbeat.msg` | Liveness + status  | 5 Hz |

Note: for  `/holosoma/holosoma_executed_cmd`, the policy backend fills all 29 values. The split-body backend fills the 14 arm joints (indices 15–28) and zeros the rest.

## Internal structure
```bash
CmdSMPLH ─▶ retargeter ─┐
                        ├─CmdDense─▶ holosoma_node (WBT) ─▶ G1   (holosoma policy)
external publisher ─────┘

[...]

PR #128, opened 2026-06-20
Unified policy service node (Variant A: reuse Ros2Input)

kingb:
## Summary

Adds a unified ROS2 **policy service node** that runs *any* holosoma policy (locomotion, WBT, …) and supports runtime policy swapping — the `run_policy.py` experience, served over ROS2. Builds on #124, reusing its `holosoma_service` package, `unitree_mp` multiprocess proxy, `TargetSource`, and `_on_command_sent` feedback hook.

**This is Variant A** (reuse `Ros2Input`): the node touches no shared input code — loco velocity/state ride the existing `Ros2Input`; the node only adds a WBT dense-target bridge and a policy-agnostic feedback publisher. A companion PR proposes Variant B (single node owns all I/O via an `injected` input source) for comparison.

> Base branch is `dev/tomasz/tracker_service` (#124), which this depends on — not `main`.

## What's here

- **`service_node.py`** — builds the policy via `_select_policy_class`/`DualModePolicy` (no extension-only `policy_type`), routes the Unitree SDK through `unitree_mp` (CycloneDDS in a child process, isolated from this process's rclpy), strips ROS args before tyro so `ros2 launch`/`run` work.
- **`feedback.py`** — `PolicyFeedbackPublisher`: policy-agnostic `executed_cmd` + `heartbeat`, driven by the per-tick `_on_command_sent` hook; each node spun on its own `SingleThreadedExecutor`.
- **`target_bridge.py`** — `DenseTargetBridge`: WBT-only `CmdDense` → `TargetSource`, attached only to policies exposing `_target_source` (loco untouched).
- **`teleop_with_loco_policy.launch.py`** — fully parameterized (`velocity_input`/`state_input`/`interface`/`domain_id`/`node_name`) for composition into a parent launch.
- Two small fixes to shared `holosoma_inference`: ROS2 `"kill"` command (parity with joystick L1+R1), and a `hasattr` guard in `dual_mode` so the SWITCH_MODE injection doesn't crash on providers without `_mapping` (ros2/injected).

## Testing

Validated **walking on a real G1** (ROS2 Jazzy) via the `ros2 launch` path: velocity over `/cmd_vel`, state on the native G1 joystick, `executed_cmd` at 50 Hz, `heartbeat` status reflecting the active policy.

### DIFF
--- a/src/holosoma_inference/holosoma_inference_service/holosoma_service/launch/teleop_with_loco_policy.launch.py
+++ b/src/holosoma_inference/holosoma_inference_service/holosoma_service/launch/teleop_with_loco_policy.launch.py
@@ -1,0 +1,101 @@
+"""Live teleop -> locomotion policy, served over ROS2 (Variant A).
+
+    /cmd_vel (Twist) ──────────────▶ policy_service_node ─▶ robot
+    state (joystick or ROS2 String) ─┘  (publishes executed_cmd + heartbeat)
+
+``policy_service_node`` builds the policy the same way ``run_policy.py`` does
+(no extension-only ``policy_type``) and runs it as a ROS2 service. The Unitree
+SDK is run through the ``unitree_mp`` multiprocess proxy automatically (the node
+forces it), so the SDK's CycloneDDS is isolated from this process's rclpy.
+
+Fully parameterized for composition into a larger launch: every knob is a
+``DeclareLaunchArgument`` so a parent launch can override it.
+
+Default config below: velocity over ROS2 ``/cmd_vel`` + state on the native G1
+joystick (so the joystick L1+R1 kill is available). A velocity publisher drives
+motion, e.g. ``demo_scripts/ros2_velocity_publisher.py --pattern shuttle``.
+
+Usage:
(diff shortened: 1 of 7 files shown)


kingb:
Companion alternative: #129 (Variant B — single node owns all I/O). These are two designs for the same goal; pick one. Both build on #124 and were validated walking on a real G1.

src/holosoma_inference/holosoma_inference_service/holosoma_service/launch/teleop_with_loco_policy.launch.py
```diff
@@ -0,0 +1,101 @@
...
+                "domain_id",
+                default_value="0",
+                description="DDS domain ID for the robot SDK communication.",
+            ),
+            DeclareLaunchArgument(
+                "node_name",
+                default_value="policy_service",
+                description="ROS2 node name (override when composing multiple instances).",
+            ),
+            Node(
```
Kedaro, line 80: My understanding is that service_node removes the need for holsoma_node, is that right? Or at least we can use the same entry point for WBT and loco? Do we still need separate launch files?

kingb:
We went with implementation B #129

PR #134, opened 2026-06-25
Add no secondary secondary policy flag

Kedaro:


src/holosoma_inference/holosoma_inference_service/holosoma_service/holosoma_service/policy_control/service_node.py
```diff
@@ -250,11 +250,25 @@ def main() -> None:
     # Under `ros2 launch` / `ros2 run`, argv carries ROS args (e.g.
     # "--ros-args -r __node:=..."). tyro would reject those, so strip them
     # before parsing this node's own CLI; rclpy.init() consumes them separately.
+    import argparse
```
tomasz-lewicki, line 253: Can we use `tyro` over `argparse` for that?

tomasz-lewicki (approved):
LGTM but can we use `tyro`?

src/holosoma_inference/holosoma_inference_service/holosoma_service/holosoma_service/policy_control/service_node.py
Kedaro: According to my agent;


The short answer is: not cleanly, no — and run_policy.py uses the same argparse pre-parse for the same reason.

The issue is that secondary: InferenceConfig | None = None gets a non-None default from the preset (e.g. g1-29dof-wbt bakes in _g1_safety_secondary). When tyro sees a subcommand default that already populates secondary with a full InferenceConfig, it generates nested subcommand groups for all the secondary's fields. There's no tyro CLI syntax to say "actually, set this whole nested struct back to None" — tyro's Optional handling works at the type level (it would need a --secondary:None subcommand variant), but subcommand_type_from_defaults doesn't generate that toggle because the defaults already provide a concrete value.

This is exactly why run_policy.py does the same thing — see line 181:

pre.add_argument("--secondary", default=None, help="Set to 'none' to disable dual-mode.")

PR #138, opened 2026-07-01
Add Integration Tests for the holosoma service API

tomasz-lewicki:
Adds a ROS2-native contract test for the service merged in #124. 

The PR tests the service's public surface ROS2 topic API.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -102,3 +102,20 @@ jobs:
           source scripts/source_isaacsim_setup.sh
           pip install -e src/holosoma_inference
           pytest -s src/holosoma_inference/
+
+  service-integration-tests:
+    # The service's contract is its ROS2 topic API, so these run in a real ROS2
+    # Jazzy workspace (the inference-tests env above has no rclpy/colcon and
+    # would only skip them). Builds the minimal test image and runs it.
+    name: Run Service Integration Tests (ROS2)
+    runs-on: codebuild-holosoma-cpu-build-${{ github.run_id }}-${{ github.run_attempt }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Build service test image
+        run: |
+          docker build \
+            -f src/holosoma_inference/holosoma_inference_service/docker/Dockerfile.test \
+            -t holosoma-service-test .
+      - name: Run service integration tests
+        run: docker run --rm holosoma-service-test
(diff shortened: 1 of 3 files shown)


kingb (commented):
Reviewed the diff against the service code on the branch and confirmed both tests actually run in CI (`collected 2 items`, both `PASSED`, `2 passed in 4.03s` in the new Run Service Integration Tests (ROS2) job) — not silently skipped. Nice, genuinely-real contract test. Three points worth raising, all minor / non-blocking:

**1. Negative test doesn't isolate the `valid` flag** (`test_retargeter_service_api.py`, `test_invalid_smplh_publishes_nothing` + `_make_smplh`)

The node guard is `if not msg.valid or not msg.joint_poses: return`. `_make_smplh(valid=False)` returns a message with *both* `valid=False` **and** empty `joint_poses`, so the test would still pass if the `valid` check were deleted — the empty-poses branch alone suppresses output. If the goal is to pin the `valid=False` contract specifically, populate `joint_poses` on the invalid frame so only the `valid` flag drives the skip.

**2. Unpinned numpy drifts across the codegen/runtime boundary** (`Dockerfile.test`, pip install block)

The Dockerfile deliberately builds messages with the system interpreter first because "the rosidl C++ typesupport generator segfaults if a newer venv numpy shadows the ROS image's numpy during msg codegen." But the `--system-site-packages` venv then pip-installs an unpinned `numpy`, and in this CI run that resolved to **numpy 2.5.0, replacing the ROS system's 1.26.4** — the very numpy that `holosoma_msgs`/rclpy C-extensions were built against. It passes today, but nothing locks that in against a future numpy release introducing an ABI mismatch at test *runtime*. Suggest pinning at least `numpy` (ideally to the jazzy system version, 1.26.4).

**3. Asymmetric pump timeout** (`pipeline` fixture / negative test)

The negative case pumps for `timeout_s=1.5` vs. `5.0` for the positive case. Discovery is already established by the preceding positive pump, so this is likely fine, but it's a thinner margin under CI load. Flag only if it ever flakes.

None of these block merge — (1) is the only one I'd nudge on, and it's a precision improvement rather than a correctness bug.

kingb (dismissed):
LGTM

asetapen (dismissed):
Great template to build upon - thanks for doing the first test @tomasz-lewicki (it's always the hardest).

tomasz-lewicki:
Thanks @kingb, addressed in 8f57f0d:

1. `_make_smplh` now populates `joint_poses` even when `valid=False`, so the invalid frame differs only in the flag and the negative test actually pins the `valid` contract (not the empty-poses branch).
2. Pinned `numpy==1.26.4` (jazzy system version) in `Dockerfile.test` 
3. Left the asymmetric pump timeout as-is since discovery is already established by the positive pump — will bump it if it ever flakes.

asetapen (approved):

PR #146, opened 2026-07-03
reproducible keyed domain-randomization sampling

juan-g-bonilla:
### Summary
Make domain randomization **reproducible and order-independent**: a single validated
`DistributionSpec` plus a keyed, counter-based `TermSampler`, so the same `(term, env, episode)`
draws the same value regardless of term set/order, `num_envs`, or which env subset resets. The
sampler is threaded through the randomization manager and every robot-DR term, and the IsaacSim
physics-DR writers are moved off IsaacLab's `mdp.randomize_rigid_body_*` onto project-owned writers
that honour the project's `[lo, hi]` bounds convention.

> ⚠️ **Not just plumbing — this changes DR *results* and backend *support* for the shipped presets
> `loco/g1`, `loco/t1`, `wbt/g1`.** The changes are individually small and (we believe) more correct,
> but **no seeded run reproduces a pre-PR run.** Treat existing checkpoints, reward curves, and
> regression fixtures as invalidated — re-baseline, don't reproduce.

## Behavior changes

All verified against the `main…02_reproducible_dr_sampling` diff and the three shipped presets.

### Different dynamics (a seeded run behaves differently, not just draws different numbers)

- **Mass DR now rescales body inertia** (`randomize_mass_startup`, `recompute_inertia` default
  **True**). Before, a mass change left inertia as authored (MuJoCo-Warp) or recomputed it from
  collision geometry (IsaacGym); now every backend scales the *authored* inertia by the mass ratio
  `m_after/m_before`. Opt out with `recompute_inertia=False`.
- **Base-CoM DR on IsaacGym no longer recomputes inertia from geometry** (`randomize_base_com_startup`;
  enabled in `loco/g1`, `wbt/g1`). A CoM offset now leaves the inertia tensor unchanged, matching the
  other backends. No opt-out.
- **IsaacSim friction DR no longer forces restitution to 0** (`randomize_friction_startup`; `loco/g1`,
  `loco/t1`). Before, the friction term silently zeroed restitution on every robot shape; now it
  leaves restitution at its authored value.
- **IsaacSim material DR now clamps dynamic friction ≤ static friction** (was an independent draw that
  could exceed static). Lowers the realized dynamic-friction on `loco/g1`/`loco/t1`; caps dynamic at
  the static draw for `wbt/g1` robot material.
- **Friction DR is quantized into `num_buckets` discrete values (default 64) on all backends** — this
  bounds the count of unique PhysX materials. MuJoCo-Warp, previously a continuous per-geom draw, is
  now a 64-value staircase by default; pass `num_buckets=None` for a continuous marginal.

### Newly supported

- **Classic (CPU) MuJoCo now runs robot mass / friction / base-CoM DR.** `mujoco_backend` defaults to
  `CLASSIC`; on `main` these terms raised `RandomizerNotSupportedError` on the classic backend, so a
  default-backend `loco/g1`/`loco/t1` run went from "DR errors at setup" to "DR applies".

### Reproducibility (no fixed seed reproduces a pre-PR run)

- **DR is drawn from the keyed sampler**, not the global RNG. Per-term distributions are unchanged and
  draws are now independent of term order / `num_envs` / reset subset, but the exact per-seed values
  differ from `main`. Affects every preset.
- **DR no longer consumes the global torch/numpy RNG stream.** Any *other* stochastic code sharing
  that stream (reset/observation/action noise, policy exploration) now sees a different sequence — a
  **run-wide** shift, not confined to DR values.
- **Per-env episode keying** re-keys reset/step-stage DR by each env's episode index (subset-stable);
  does not change *when* or *how often* any term fires.

### Latent (no shipped preset triggers these today)

- A DR run with **no seed now raises** (`TermSampler` requires one; `TrainingConfig.seed` defaults to
  42, so shipped presets never hit it).
- A `gaussian` spec is now a **truncated** normal on `[lo, hi]` (was an unbounded `randn·std+mean`);
  `log_uniform` bounds are validated (finite, strictly positive) at construction.
- Term-name collisions across DR term classes now **raise** instead of silently reusing an instance.

### Risk
Moderate. **Config-driven DR needs no API change** (the manager injects the sampler) and every
preset's ranges are preserved — but the PR is **not** result-preserving (see above). Library-level
break: custom `RandomizationTermBase` subclasses and direct DR-term calls now take a keyword-only
`sampler: TermSampler` (Migration Guide §1).

### 1 Custom DR terms / direct term calls gain a `sampler`
`setup`/`reset`/`step` and every term function now require a keyword-only `sampler: TermSampler`
(bind with `TermSampler.bind(seed, term_name, stage, episode_count)`). **Config-only users need no
change** — the manager injects it. Range params are typed `DistributionLike`: the same `[lo, hi]`
pair, or `{kind, low, high, mean, std}` for gaussian / log_uniform (bounds validated at construction).

### 2 Module rename
`simulator.mujoco.backends.warp_randomization` → `simulator.mujoco.backends.randomization`.

### 3 IsaacSim physics-DR writers take specs + `sampler`
The writers in `simulator/isaacsim/events.py` no longer route through IsaacLab's
`mdp.randomize_rigid_body_*`. `randomize_body_com` / `randomize_rigid_body_inertia` take a
`Sequence[DistributionSpec]`; `randomize_rigid_body_mass` / `randomize_rigid_body_material` take a
`DistributionLike` (or per-channel config); all take a keyword-only `sampler`. The old `distribution=`
/ `*_distribution_params` arguments are removed.

### 4 Re-baseline, don't reproduce
DR results change and no seed reproduces a pre-PR run, so any pre-PR checkpoint / reward curve /
regression fixture is not comparable value-for-value. `recompute_inertia=False` restores the old mass
behaviour; the CoM-inertia, restitution, and friction-quantization changes are intended fixes with no
opt-out — re-baseline against them.

### DIFF
--- a/src/holosoma/holosoma/managers/randomization/terms/_shared.py
+++ b/src/holosoma/holosoma/managers/randomization/terms/_shared.py
@@ -1,0 +1,20 @@
+"""Small helpers shared by the randomization term modules (locomotion + objects)."""
+
+from __future__ import annotations
+
+from typing import Any, Sequence
+
+import torch
+
+# PhysX caps the number of unique materials per scene, so IsaacSim material DR fills a fixed bucket
+# table and assigns each shape a bucket. 64 buckets keep the quantile staircase close to continuous.
+_MATERIAL_NUM_BUCKETS = 64
+
+
+def _ensure_env_ids_tensor(env: Any, env_ids: torch.Tensor | Sequence[int] | None) -> torch.Tensor:
+    """Convert environment indices to a tensor on the correct device."""
+    if env_ids is None:
+        return torch.arange(env.num_envs, device=env.device, dtype=torch.long)
+    if isinstance(env_ids, torch.Tensor):
+        return env_ids.to(device=env.device, dtype=torch.long)
+    return torch.as_tensor(list(env_ids), device=env.device, dtype=torch.long)
--- a/src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
+++ b/src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
@@ -1,20 +1,27 @@
-"""MuJoCo Warp domain randomization utilities.
+"""MuJoCo domain randomization utilities (both backends).
 
 Adapted from mjlab (Apache 2.0): https://github.com/mujocolab/mjlab/blob/main/src/mjlab/sim/randomization.py
 See THIRD_PARTY_LICENSES for full license text.
 
-Provides functions for randomizing model fields across batched environments
-in GPU-accelerated MuJoCo Warp simulations.
+Provides ``randomize_field`` for randomizing model fields, which works on BOTH MuJoCo
+backends: the WarpBackend (GPU, vectorized across the per-world-expanded model via the
+WarpBridge) and the single-env ClassicBackend (CPU, writing the one ``mujoco.MjModel`` field
+in place). ``expand_model_fields`` and its Warp kernel are WarpBackend-only — they tile a
+single-world model across environments and are a no-op for the single-env ClassicBackend.
 """
 
 from __future__ import annotations
 
-from typing import Any, Callable, Literal, cast
+from typing import TYPE_CHECKING, Any, Callable, Literal, cast
 
 import mujoco
-import mujoco_warp as mjwarp
 import torch
 import warp as wp
+
+from holosoma.utils.sampler import DistributionLike, DistributionSpec, TermSampler, quantiles
+
+if TYPE_CHECKING:
+    import mujoco_warp as mjwarp
 
 
 @wp.kernel(module="unique")
@@ -81,7 +88,11 @@ def expand_model_fields(
         return dst
 
     for field in model.__dataclass_fields__:
-        if field in fields_to_expand:
+        # Skip fields already expanded: re-tiling REALLOCATES the array (setattr below), which
+        # would invalidate a captured CUDA step-graph that still points at the old array. Making
+        # this idempotent lets the runtime static-move path call it safely (it's a no-op once the
+        # field was expanded at setup); only the first expansion per field allocates.
+        if field in fields_to_expand and field not in model._expanded_fields:  # type: ignore[attr-defined]
             array = getattr(model, field)
             setattr(model, field, tile(array))
             # Register this field as expanded for validation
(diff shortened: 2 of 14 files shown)


src/holosoma/holosoma/managers/randomization/terms/_shared.py
```diff
@@ -0,0 +1,20 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Sequence
+
+import torch
+
+# PhysX caps the number of unique materials per scene, so IsaacSim material DR fills a fixed bucket
+# table and assigns each shape a bucket. 64 buckets keep the quantile staircase close to continuous.
+_MATERIAL_NUM_BUCKETS = 64
```
samuelgundry, line 11: Did you hit this error when randomizing objects OR an existing robot task/env?

src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
```diff
@@ -144,32 +155,75 @@ def resolve_entity_ids(mj_model: mujoco.MjModel, names: list[str], entity_type:
...
+
+def scale_inertia_by_mass_ratio(
+    simulator: Any, body_ids: torch.Tensor, mass_before: torch.Tensor, *, min_mass: float = 1e-6
+) -> None:
+    """Scale ``body_inertia`` rows for ``body_ids`` by each body's CURRENT/``mass_before`` ratio.
+
+    The MuJoCo half of the cross-backend ``recompute_inertia`` contract: after a mass write, multiply
+    each body's diagonal inertia (``body_inertia`` is ``[nbody, 3]`` principal moments) by the factor
+    its mass changed by, so a uniform-density mass change scales inertia consistently with the Isaac
+    backends. Multiplicative => commutes with a separate inertia-shape DR term (order-independent).
```
samuelgundry, line 179: This will change the existing behavior for robot envs/tasks in Mujoco when `recompute_inertia=True`, not just obiects right? Even though it's more accurate, might be worth noting in the PR description. I don't think we have nightly Mujoco-based CI results to compare against?

samuelgundry (dismissed):
Looks good. A minor Q and just a comment on noting the recompute_inertia changes in the PR description given it will subtlety change robot physics.

src/holosoma/holosoma/managers/randomization/terms/_shared.py
juan-g-bonilla: no, I didn't run into it because I didn't launch a big DR campaign, but theoretically with 10s of shapes per environment and thousands of environments we would hit the limit (+ there might be performance penalties if we keep building up materials infinitely). This strategy is basically lifted from the bucket approach used by isaaclab to prevent this problem. I just pushed a new version of this PR that removes this baked in magic number and adds `num_buckets` as a parameter that can be tuned by the user, including using `num_buckets=None` to not use buckets at all.

src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
juan-g-bonilla: Good catch. claude and I went through the code again trying to capture all the behavior changes like this. See the updated PR description when you have a second.

samuelgundry (approved):
Thanks for adding the behavior changes to the description.

PR #147, opened 2026-07-03
backend-agnostic scene objects + link_physics

juan-g-bonilla:
# PR #147 — `feat: backend-agnostic scene objects + shared robot link_physics`

## Summary

Turns the single, IsaacSim-only rigid-object feature into a general **backend-agnostic scene-asset
system** and adds the shared robot **`link_physics`** mapping. Declare any number of free and static
scene bodies in one top-level `scene:` config; spawn them on all four backends (MuJoCo Classic CPU,
MuJoCo Warp GPU, IsaacGym, IsaacSim); address them through one unified world-frame actor-state API;
observe, reset, and pose-jitter them; and map a robot's link physics with the same `PhysicsConfig`
objects use. This is the breaking PR of the stack. *Domain-randomizing* those objects is PR #148.

## ⚠️ Not result-preserving — re-baseline, don't reproduce

Beyond the config/API break, this PR changes **physics results** for shipped presets. No seed
reproduces a pre-PR run on the affected paths:

- **MuJoCo base angular-velocity observation changes for every robot policy.** Both MuJoCo root-state
  views now rotate the freejoint angular velocity body-local→world on read (and inverse on write)
  (`mujoco/mjw_views.py:566`, `mujoco/tensor_views.py:296`). Previously `robot_root_states[:,10:13]`
  was raw body-local, so `base_ang_vel = quat_rotate_inverse(base_quat, ω)` applied a spurious second
  rotation (correct only when `base_quat≈identity`). This is now correct and matches IsaacGym/IsaacSim
  — but `base_ang_vel` and every reward/command term derived from it shifts for any policy trained or
  evaluated on `mujoco`/`mjwarp`. Sim2sim-relevant.
- **MuJoCo Warp runtime/startup DR now actually affects dynamics.** `WarpBackend.__init__` pre-expands
  the per-world model fields (`body_mass`/`body_inertia`/`geom_friction`/`dof_damping`/… ) *before*
  CUDA-graph capture (`mujoco/backends/warp_backend.py:250`). Before, DR writes hit arrays reallocated
  after capture — orphaned from the step graph, so mass/friction/inertia/CoM DR read back changed but
  never moved the physics. Now `loco/g1`, `loco/t1`, `wbt/g1` startup DR is live on `mjwarp`.
- **MuJoCo `num_bodies` / body-indexed tensors re-based.** `num_bodies` goes from `root_model.nbody`
  (world included) to `len(body_names)` (world excluded), and `contact_forces` / `_rigid_body_*` are
  gathered so row *k* = `body_names[k]` (`mujoco/mujoco.py:443`). This fixes a WBT body-tracking
  off-by-one on MuJoCo (tracked-body rows were shifted by the world body); the *shape* of
  `contact_forces`/`_rigid_body_*` and the *value* of `num_bodies` both change.
- **WBT-with-object episode horizon corrected 20 s → 10 s** (`config_values/wbt/g1/experiment.py:150`,
  `178`). On `main`/base the `_w_object` presets set `env_spacing=0.0` via a full-field
  `simulator=replace(simulator.isaacsim, …)` rebuilt from the *raw* IsaacSim preset — which carries no
  episode override and so defaulted to `SimEngineConfig`'s 20.0, collaterally clobbering the base
  `g1_29dof_wbt` override of `max_episode_length_s=10.0`. This PR moves `env_spacing=0.0` into the scene
  preset and drops the `simulator=` override, so the object presets now correctly **inherit 10.0** —
  matching the sibling non-object `g1_29dof_wbt` preset and the original unsplit branch. **Not a
  regression** (the intended horizon is 10 s), but it does change the horizon vs `main`, so re-baseline
  any `g1_29dof_wbt_w_object` / `g1_29dof_wbt_fast_sac_w_object` checkpoint trained at the accidental 20 s.
- **IMU gyro now body-frame in the sim2sim bridge** (`bridge/base/basic_sdk2py_bridge.py:300`).
  `_get_base_imu_data` rotates world→body (`quat_rotate_inverse(base_quat, ω)`) so the gyro matches a
  physical IMU. Any deployed policy driven through the bridge (`run_sim`, all backends) sees a
  different angular-velocity signal.
- **Object physics-DR temporarily removed from the shipped WBT-object preset.**
  `g1_29dof_wbt_randomization_w_object` drops `object_state_dr_at_setup` (object material/mass/inertia
  DR), making it byte-identical to the base preset within this PR
  (`config_values/wbt/g1/randomization.py:104`). The object runs un-randomized here; PR #148 re-adds it
  cross-backend in `terms/objects.py`.

### Different dynamics (existing runs)
- **MuJoCo base ang-vel world-frame** — see ⚠️ (`mujoco/mjw_views.py:566`, `tensor_views.py:296`).
- **MuJoCo Warp DR pre-capture expansion** — see ⚠️ (`mujoco/backends/warp_backend.py:250`).
- **MuJoCo `num_bodies` / body-indexed tensors re-based** — see ⚠️ (`mujoco/mujoco.py:443`).
- **WBT-object episode corrected 20 s → 10 s** — see ⚠️ (`config_values/wbt/g1/experiment.py:150`).
- **IMU gyro body-frame in bridge** — see ⚠️ (`bridge/base/basic_sdk2py_bridge.py:300`).
- **World-frame actor-state contract standardized across backends** for `get_actor_initial_poses` /
  `set_actor_states` / `get_actor_states` — env origins baked in at spawn; callers pass world coords
  and must not add `env_origins` or branch on simulator type (`base_simulator.py`,
  `isaacgym/isaacgym.py`, `isaacsim/state_adapter.py`). Numerically a no-op for shipped presets (WBT
  uses `env_spacing=0.0`); an API-semantics change for `env_spacing>0`.
- **MuJoCo `set_actor_states_by_index` writes each env's own freejoint slice** on live backend storage
  (was single-env, hardcoded `qpos[0:3]`) — fixes multi-env root-state writes on `mujoco`/`mjwarp`
  (`mujoco/mujoco.py`).

### Newly supported (errored → works, or new capability)
- **Scene `rigid_objects` spawn on IsaacGym** (were ignored entirely; `_load_scene` returned early
  unless `scene_files` was set) (`isaacgym/isaacgym.py:271`).
- **MuJoCo scene-asset loading** — `get_supported_scene_formats()` now `['xml','urdf']` (was `[]`); the
  WBT box loads on `mujoco`/`mjwarp` (`mujoco/mujoco.py`). Static (welded, jointless) bodies get a new
  kinematic-teleport placement path on both MuJoCo backends.
- **IsaacSim scene objects accept URDF** — `get_supported_scene_formats()` `['usd']` → `['usd','urdf']`;
  the URDF-only WBT box now spawns through the IsaacSim URDF→USD scene path (`isaacsim/isaacsim.py:528`).
- **WBT-with-object task is now backend-agnostic** — the `assert simulator == ISAACSIM`
  ("Object is only supported in IsaacSim") is gone; the object name is resolved from the scene registry
  (`get_names_by_type(INDIVIDUAL)[0]`) and its state is read via `get_actor_states` instead of indexing
  `all_root_states` (out-of-bounds on MuJoCo). WBT-object no longer errors at setup off IsaacSim
  (`managers/command/terms/wbt.py:601`, `:1023`).
- **MuJoCo object/actor access no longer rejected** — `get/set_actor_states_by_index`,
  `get_actor_indices`, `get_actor_initial_poses` accept any registered name (unknown → `KeyError`)
  (`mujoco/mujoco.py`).
- **Robot-only IsaacGym scenes register `"robot"`** — `_register_objects` runs unconditionally, so the
  actor-state API works on `loco/g1`, `loco/t1`, `wbt/g1` under IsaacGym (`isaacgym/isaacgym.py`).
- **1→N scene files** expand to per-body `RigidObject`s named `{file}_{body}` on IsaacSim/IsaacGym
  (`isaacsim/object_spawner.py`, `isaacgym/urdf_scene_loader.py`).
- **Free objects' configured initial velocities applied at startup** on IsaacGym/IsaacSim/MuJoCo
  (`linear_velocity`/`angular_velocity`; `object-managers-demo`'s `moving_box`).
- **New object managers** (opt-in, additive; no shipped preset wires them): base-frame object
  observation terms (`managers/observation/terms/objects.py`), per-episode object reset
  (`_reset_objects_callback`, `base_task.py:400`), and `jitter_object_pose_on_reset` (free-body XY/yaw
  only, never static) (`managers/randomization/terms/objects.py`).
- **Robot `link_physics` applied at spawn** — IsaacGym/IsaacSim robot links can receive
  friction/restitution/compliance + collision/mass props via `link_physics` (new
  `apply_rigid_shape_properties` / material bind). Latent for shipped robots (their per-backend
  material blocks are `None`); PhysX scalars migrate value-for-value (see Migration §5). MuJoCo applies
  `link_physics` pre-compile with no dynamics change for g1/t1 (no freejoint-damping change).
- **`create_virtual_gantry(enable=False)` no longer requires an attachment body** — a non-humanoid /
  single-link robot lacking a torso/pelvis body now boots with the gantry off instead of raising at
  setup (`shared/virtual_gantry.py:532`). Default is gantry-off, so this is the common path.
- **Headless IsaacGym `render()` no-ops** when `viewer is None`, fixing an `AttributeError` on the
  `run_sim` direct loop (`isaacgym/isaacgym.py:829`).

### Removed / fail-loud (works → errors for malformed input)
- **IsaacGym rigid-object / scene-file loading is now fail-loud** — a missing file or unsupported
  format raises `ValueError` instead of being silently skipped with a warning; multiple scene files are
  supported (was: >1 URDF scene file raised) (`isaacgym/urdf_scene_loader.py:441`).
- **IsaacSim USD spawns validate composition** — `CustomUsdFileCfg.validate_composition` defaults
  `True`; a USD that composes no renderable geometry or has materials resolving outside the subtree now
  raises at spawn (was silently invisible) (`isaacsim/spawners/from_files_cfg.py:58`). No shipped USD
  asset triggers it.
- **Physics sub-configs reject unknown fields** (`ConfigDict(extra='forbid')` on the five physics
  dataclasses) — a typo'd or now-relocated field fails at construction (`config_types/scene.py:19`).

### Reproducibility
Covered by the ⚠️ items: the MuJoCo ang-vel frame, Warp DR pre-capture expansion, `num_bodies`
re-basing, the WBT episode-length correction (20 s → intended 10 s), and the bridge IMU frame each break value-for-value
comparison against pre-PR checkpoints/baselines on the affected path.

### Latent / edge-case
- **`ObjectRegistry.resolve_indices` preserves first-occurrence order** (`dict.fromkeys`) instead of
  ascending-sorted (`torch.unique`) (`shared/object_registry.py:412`). A caller pairing
  `get_actor_initial_poses` with the by-index write path previously applied poses to the *wrong* actor
  when request order ≠ position order; now correct. Affects multi-object scenes only
  (`object-managers-demo`); single-object `wbt/g1` and empty `loco/*` are unaffected.
- **IsaacGym `contact_forces` sliced to robot bodies** (`[:, :num_bodies, :]`) — no-op for robot-only
  shipped presets; matters once scene objects add bodies (`isaacgym/isaacgym.py:658`).
- **`find_rigid_body_indice` (MuJoCo) returns a 0-based `body_names` index** (world excluded), not a raw
  MuJoCo body id — a direct caller passing it to `apply_force`/xfrc must map through `body_ids[index]`
  (`mujoco/mujoco.py`). In-tree consumers updated; the virtual-gantry force path maps through
  `body_ids` accordingly.
- **`WarpBackend` is always a class, never `None`** — an import sentinel replaces the `None` binding, so
  `isinstance(_, WarpBackend)` is a safe `False`; external code testing `WarpBackend is None` for
  CPU-only installs must switch to the `WARP_AVAILABLE` flag (`mujoco/backends/__init__.py:32`).
- **IsaacSim WBT object spawns with 10× higher default damping** (0.01→0.1) — the object is now a
  generic scene `RigidObjectConfig(physics=None)` that falls back to `PhysXPhysicsConfig()` defaults
  (`isaacsim/converters.py:119`). Damping is a persistent spawn property, but the WBT command term
  overwrites the object pose+velocity from motion data every reset, so the steady-state effect is
  largely masked; **verify on re-baseline** if WBT-object settling behavior matters.

## Migration Guide

*(The stack-wide guide §3 covers most of this. Below is verified against the tip; two entries there are
**stale** — flagged inline.)*

### 1 · Scene objects: `RobotConfig.object` → top-level `scene=`
`RobotConfig.object` and the `ObjectConfig` type are **removed**. Declare scene objects via a new
top-level `scene:` config — a field on `EnvConfig`, `ExperimentConfig`, `FullSimConfig`, `RunSimConfig`
(`FullSimConfig.scene` has a `default_factory`, so constructing without `scene=` still works).
`SimulatorInitConfig.scene` is also removed and the simulator presets no longer carry
`scene=SceneConfig(replicate_physics=True)`.

```python
# BEFORE — object hung off the robot config
ExperimentConfig(robot=replace(robot.g1_29dof_w_object,
    object=replace(robot.g1_29dof_w_object.object, object_urdf_path=".../objects_largebox.urdf")))
# AFTER — scene is its own top-level config
ExperimentConfig(robot=robot.g1_29dof_w_object,
    scene=scene.g1_29dof_wbt_object_scene)   # from holosoma.config_values import scene
```
CLI: `--simulator.scene.env_spacing=0.0` → `--scene.env_spacing=0.0`. Asset moved:
`.../whole_body_tracking/objects_largebox.urdf` → `data/scene_objects/boxes/large_box.urdf`.

### 2 · Scene objects/files are name-keyed dicts, not lists
`SceneConfig.rigid_objects` / `scene_files` are `dict[str, ...]` keyed by actor name (the registry key);
the `name` field is removed. Both default to `{}` (were `None`). A model validator rejects a key shared
between `rigid_objects` and `scene_files`.

```python
# BEFORE
SceneConfig(rigid_objects=[RigidObjectConfig(name="box", urdf_path="box.urdf"),
                           RigidObjectConfig(name="post", urdf_path="post.urdf",
                                             physics=PhysicsConfig(kinematic_enabled=True))])
# AFTER — dict keyed by name; fixed=True for static
SceneConfig(rigid_objects={"box":  RigidObjectConfig(urdf_file="box.urdf"),
                           "post": RigidObjectConfig(urdf_file="post.urdf", fixed=True)})
```

### 3 · `RigidObjectConfig` / `SceneFileConfig` field changes
- **`RigidObjectConfig`**: `urdf_path`→`urdf_file`, `usd_path`→`usd_file`; added `xml_file`.
- ⚠️ **STALE in guide §3.3**: the `*_path → *_file` rename applies **only to `RigidObjectConfig`**.
  `SceneFileConfig` **keeps** `usd_path`/`urdf_path` and merely *adds* `xml_path`
  (`config_types/scene.py:226-232`); the shipped `object-managers-demo` preset constructs
  `SceneFileConfig(xml_path=…, urdf_path=…, usd_path=…)`. Passing `urdf_file=` to `SceneFileConfig`
  raises `TypeError`. The two sibling types are asymmetric by design.
- Added to `RigidObjectConfig`: `asset_root`, `linear_velocity`, `angular_velocity`, `fixed`.
- **Removed: `scale`** — no replacement; author the asset at its intended size.
- Removed: `PhysicsConfig.kinematic_enabled` — use `fixed=True`.
- Removed: `SceneFileConfig.urdf_settings` / the `URDFSettings` type + its URDF-transform validator.
- Removed methods: `SceneFileConfig.get_asset_path()` / `has_format()` — format selection goes through
  `holosoma.simulator.shared.asset_format.select_asset_format(obj, supported_formats)`.
- New validators: non-zero velocity on `fixed=True` rejected; physics sub-configs `extra='forbid'`.
- `IsaacSimPhysicsConfig` gained `contact_offset` / `rest_offset` / `torsional_patch_radius` (+ an
  offset-ordering validator); `ObjectPatternConfig` gained `fixed` / `position_offset` overrides.

### 4 · Physics config classes moved + split per backend
`SceneConfig`, `SceneFileConfig`, `RigidObjectConfig`, `ObjectPatternConfig`, `PhysicsConfig` moved
`config_types.simulator` → `config_types.scene` (no back-compat re-export). Note `config_types.simulator`
still exists and keeps `SimEngineConfig`/`SimulatorConfig`/etc — only the scene types moved.
`PhysicsConfig` no longer bundles backend settings inline — per-backend sub-configs each name only the
channels that backend honors: `PhysXPhysicsConfig` (`physx`, shared by both Isaac backends),
`IsaacGymPhysicsConfig` (`isaacgym`), `IsaacSimPhysicsConfig` (`isaacsim`), `MujocoPhysicsConfig`
(`mujoco`).

```python
from holosoma.config_types.scene import SceneConfig, RigidObjectConfig  # was config_types.simulator
```

### 5 · Robot PhysX scalars → `RobotAssetConfig.link_physics`
The four loose PhysX scalars (`max_angular_velocity`, `max_linear_velocity`, `angular_damping`,
`linear_damping`) and `density` are **removed** from `RobotAssetConfig`. Set them via `link_physics`, a
shared `PhysicsConfig` applied to every robot link (`body_names='.*'`). `mass` is rejected on
`link_physics`. Values migrate 1:1 for shipped robots (`g1_29dof`, `t1_29dof_waist_wrist`).

```python
# BEFORE
RobotAssetConfig(..., max_angular_velocity=1000.0, max_linear_velocity=1000.0,
                 angular_damping=0.0, linear_damping=0.0, density=0.001)
# AFTER
RobotAssetConfig(..., link_physics=PhysicsConfig(
    density=0.001,  # omit if the asset authors mass
    physx=PhysXPhysicsConfig(linear_damping=0.0, angular_damping=0.0,   # explicit 0.0 — the
                             max_linear_velocity=1000.0,                 # PhysXPhysicsConfig
                             max_angular_velocity=1000.0)))              # default damping is 0.1
```

### 6 · Object observation / reset / jitter terms
New object *observation* terms (base-frame), per-episode object *reset* (`_reset_objects_callback`), and
`jitter_object_pose_on_reset` (opt-in; free-body XY/yaw only, never static) live in
`managers/{observation,randomization}/terms/objects.py`. They target scene objects via the registry;
robot-only scenes leave them a no-op. **Object physics-DR (mass/material/inertia) is NOT in this PR** —
it moves to `terms/objects.py`, cross-backend, in PR #148.

### 7 · `BaseSimulator` API
- **World-frame actor state (all backends, incl. `"robot"`)** — `get_actor_initial_poses`,
  `set_actor_states`, `get_actor_states_by_names` operate in world coords (env origins baked at spawn).
  Don't add `env_origins`; don't branch on simulator type. `get → set` round-trips unchanged.
- **`all_root_states` is a `UnifiedRootStatesView`, not a `torch.Tensor`** (replaces IsaacSim's
  `AllRootStatesProxy`). Spans all actors, indexable with `get_actor_indices`. Duck-types
  `view[idx]`, `view[idx, cols]` r/w, `.shape/.device/.dtype/.clone()` — but `isinstance(_, torch.Tensor)`,
  `.view()/.reshape()/.to()`, and `gymtorch.unwrap_tensor(...)` no longer work. Use `set_actor_states`
  for object writes; passing the view to `set_actor_root_state_tensor` writes the robot only. On MuJoCo
  the view is robot-only — index objects via `get_actor_states`, not `all_root_states`.
- **`get_actor_indices` ordering is name-major, env-minor** (all of `names[0]`'s envs, then `names[1]`'s).
  Absolute index values still differ per backend — never hardcode.
- **New/changed methods:** `set_static_body_pose(names, env_ids, poses)` (new);
  `get_actor_initial_velocities(names, env_ids)` (new — used by the object-reset callback);
  `clear_contact_forces_history(env_ids: torch.Tensor)` (was a single `int`; annotation/rename only,
  impls already took tensors); `get_supported_scene_formats() -> list[str]` (already abstract in the
  base — this PR only adds the return-type hint and makes it *load-bearing* via `select_asset_format`, so
  it's not a *newly-required* method); `MuJoCo.find_rigid_body_indice` now returns a 0-based
  `body_names` index (raw id via `sim.body_ids[index]`); MuJoCo object access no longer rejected.
- **Custom backend subclasses:** scene backends must implement `_collect_spawned_actors()`;
  `IMujocoBackend` subclasses must implement `compute_contact_forces()`, `get_actor_state()`,
  `set_actor_state()`, `set_static_body_world_pose()` (old `refresh_sim_tensors(contact_history_tensor)`
  / `create_force_view(num_bodies)` removed). `ObjectRegistry.register_object` gains
  `initial_velocities`; `registry.objects` entries are now a 6-field `RegisteredObject` NamedTuple (was
  a 5-tuple) — tuple-unpacking consumers break.

### 8 · Removed / renamed modules
| Old | Replacement |
|---|---|
| `simulator.isaacsim.usd_file_loader` | gone — use `simulator.isaacsim.object_spawner` |
| `simulator.isaacsim.usd_physics_utils` | gone (dead after consolidation) |
| `simulator.isaacsim.path_utils` | gone — use `holosoma.utils.path.resolve_asset_path` |
| `simulator.isaacsim.registry_utils` | gone (dead after consolidation) |
| `AllRootStatesProxy` (`simulator.isaacsim.proxy_utils`) | `UnifiedRootStatesView` (`simulator.shared.root_states_view`) |

`RootStatesProxy` (the robot-only wxyz↔xyzw wrapper) in `proxy_utils.py` is unchanged.


src/holosoma/holosoma/bridge/base/basic_sdk2py_bridge.py
```diff
@@ -292,7 +293,12 @@ def _get_base_imu_data(self):
                 - acceleration: linear acceleration [ax, ay, az] (3 elements)
         """
         quat_holosoma = self.simulator.robot_root_states[0, 3:7]  # [x, y, z, w]
-        gyro = self.simulator.robot_root_states[0, 10:13]  # Angular velocity
+        # robot_root_states[:, 10:13] is WORLD-frame angular velocity on every backend (the
+        # unified contract). A physical IMU gyro reports angular velocity in the BODY frame,
+        # so rotate world -> body using the base orientation. This is backend-agnostic and
+        # keeps the gyro body-frame on MuJoCo, IsaacGym, and IsaacSim alike.
+        ang_vel_world = self.simulator.robot_root_states[0, 10:13]
+        gyro = quat_rotate_inverse(quat_holosoma.unsqueeze(0), ang_vel_world.unsqueeze(0), w_last=True).squeeze(0)
```
samuelgundry, line 301: Let's notify downstream consumers this bug and fix.

src/holosoma/holosoma/managers/command/terms/wbt.py
```diff
@@ -585,13 +596,12 @@ def setup(self) -> None:
 
         # 3. get the name of the object, or indices of the object
         if self.motion.has_object:
-            # cache the object_index_in_simulator
-            self.object_name = "object"  # hardcoded object name
-            self.object_indices_in_simulator = self._env.simulator.get_actor_indices(self.object_name, env_ids=None)
```
samuelgundry, line 590: I don't have a sense of how many consumers could depend on these members, whether in holosoma or downstream users.  Any ideas on how we can depcrate + warning while retaining these members for backwards compatibility?

src/holosoma/holosoma/managers/randomization/terms/objects.py
```diff
@@ -0,0 +1,187 @@
...
+    if idx.numel() == 0:
+        return
+
+    # Only INDIVIDUAL (free) bodies are jitterable; never write a static body's pose. An explicitly
+    # requested static name is a config mistake — warn instead of silently skipping it.
+    static = set(env.simulator.object_registry.get_names_by_type(ObjectType.SCENE))
+    resolved = _resolve_object_names(env, object_names)
+    dropped_static = [n for n in resolved if n in static]
+    if dropped_static and object_names is not None:
+        logger.warning(f"pose jitter skips static (SCENE) bodies: {dropped_static}")
```
samuelgundry, line 149: Not to change but it'd be nice to explore how to "pre-validate" these kinds of misconfigurations.

src/holosoma/holosoma/simulator/isaacsim/events.py
```diff
@@ -1,161 +1,241 @@
+"""Project-owned IsaacSim physics randomization writers (mass / CoM / inertia / material).
+
+These deliberately do NOT route through IsaacLab's ``mdp.randomize_rigid_body_*`` events. IsaacLab's
+samplers read a gaussian's two parameters as ``(mean, std)``, which conflicts with the project's
+``[lo, hi]`` BOUNDS convention and silently produces a wrong distribution. By owning the write here
+and sampling through the shared keyed :meth:`holosoma.utils.sampler.TermSampler.draw`, a config means exactly
+the same thing on IsaacSim as on MuJoCo and IsaacGym — including gaussian and log_uniform.
```
samuelgundry, line 7: We should aspire to user-facing guides as docs.

samuelgundry (commented):
Epic effort, setting up for generalizing Holosoma further. Awesome stuff.  The unified proxy is a good refactor. I look forward to seeing more envs/tasks leveraging object interactions.

I did not review test codes (there's a lot) but I do want to know the level of regression testing on e.g, CI/training results and include this in the PR description. This will help traceability.

src/holosoma/holosoma/bridge/base/basic_sdk2py_bridge.py
juan-g-bonilla: Good point, I expanded the PR description to include all the breaking changes (even if fixes) including this.

src/holosoma/holosoma/managers/command/terms/wbt.py
juan-g-bonilla: This is a larger philosophy question on whether we need to respect old APIs through a deprecation process, or whether we're in v0 phase where we can change APIs. I think the latter would make the code less clean, slow down development, and make us more afraid of larger refactors. The former would be nice for our users, but most of them would ignore any deprecation warnings, then have the phase the same problem when the full error is raised when the deprecation cycle ends, and have their agents fix it anyway.

src/holosoma/holosoma/simulator/isaacsim/events.py
juan-g-bonilla: Meaning docstrings should be geared towards tool users, and not developers? Or tool users (people writing Python code with holosoma)? To me this paragraph reads like dev notes: things useful for developers of holosoma and people who use these functions when importing holosoma, but not necessarily end users.

samuelgundry (approved):
Awesome effort. Thanks for updating the behavior changes in the PR description. Good to notify folks. These changes set up a foundation, yay!

PR #149, opened 2026-07-03
holosoma core: convert to pyproject-only + uv_build backend + README

clayrosenthal:
*Description of changes:*

Converting holosoma core to use `uv-build` backend and pyproject.toml for metadata, including uploading the README.md properly.


src/holosoma/pyproject.toml
```diff
@@ -1,11 +1,24 @@
 [build-system]
-requires = ["setuptools>=45", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.9.9,<0.10"]
```
tomasz-lewicki, line 2: Can we still build with setuptools?

tomasz-lewicki (approved):

src/holosoma/pyproject.toml
clayrosenthal: Do we need setuptools for anything or you just like it better?

PR #152, opened 2026-07-03
ci(docker-build): gate :latest behind mypy+pytest; promote on push-to-main

clayrosenthal:
## Problem

Two bugs in `docker-build.yaml`, both hit by the recent SDK-stub-fix merge (#151):

1. **push-to-main never tags `:latest`.** `:latest` was only appended when `inputs.tag_latest == 'true'` — a `workflow_dispatch`-only input defaulting to `false`. A merge to main triggers the `push` path, where `tag_latest` is unset → `:latest` was never pushed. Since the ami-builder's EventBridge rule fires on an ECR **`:latest`** push, **merges never refreshed the runner AMI** (verified: 0 rule invocations after the #151 merge; no AMI build triggered).
2. **`:latest` was pushed atomically with the build**, before any test ran — so a broken image could land on `:latest`, which every CI runner *and* the AMI builder consume.

## Fix — split `build → test-gate → promote`

- **`build`** now pushes only immutable tags (`sha-<short>` + date). It **never** pushes `:latest`.
- **Test gate** (only when promoting) runs the same checks as CI, against the freshly-built `sha-` images:
  - `gate-mypy` — `mypy .` (mirrors `static-checks.yaml`)
  - `gate-sim-tests` — `isaacgym` + `isaacsim` suites (`tests/ci/<sim>_ci_tests.sh`)
  - `gate-e2e-tests` — `isaacgym` e2e (`tests/ci/isaacgym_e2e_ci_tests.sh`)
  - `gate-inference-tests` — `pytest -s src/holosoma_inference/`
- **`promote-latest`** retags `sha-<short> → :latest` in ECR via `docker buildx imagetools create` (manifest copy — no rebuild, preserves multi-arch, byte-identical to the gated image). It runs **only if every gate job succeeded** (each `needs.*.result == 'success'`), so a failed/skipped gate leaves `:latest` on the last-known-good image.

### When does `:latest` move?
`promote == true` when the run is a **push to main** *or* a **dispatch with `tag_latest=true`**. Either way it's now gated behind the full test suite. A dispatch with `tag_latest=false` still builds + pushes `sha-`/date tags but skips the gate and leaves `:latest` untouched.

### Net effect
A merge to main now auto-refreshes the runner AMI (the original intent) — but only with an image that passed mypy + the pytest suites. This closes the gap that stranded the SDK-stub-fix AMI rebuild.

## Notes
- The pre-existing `continue-on-error: true` on `build` is retained so one image failing doesn't abort the others; the gate jobs pin to `holosoma`/`holosoma-<sim>` sha images, so a failed core/sim build surfaces as a gate failure and correctly blocks promotion.
- No change to Dockerfiles, test scripts, or runner labels — only the workflow's job graph.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       tag_latest:
-        description: 'Tag as latest'
+        description: 'Promote to :latest after tests pass'
         required: true
         type: choice
         options:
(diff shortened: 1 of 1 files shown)


juan-g-bonilla (approved):
lgtm

PR #154, opened 2026-07-08
docs(service): refresh inference-service README

kingb:
Catches the service README up to the current architecture: unified `policy_service_node` (single node owns all I/O, replaces `holosoma_node`), the loco launch file, `/cmd_vel` + `/holosoma/state_input` topics (in-band kill), configurable depth input, retargeter entry-point registry, heartbeat cadence, and the ROS2 integration-test workflow. Also fixes the `simplh` typo and adds the required `preset` arg to the WBT examples.

### DIFF
--- a/src/holosoma_inference/holosoma_inference_service/README.md
+++ b/src/holosoma_inference/holosoma_inference_service/README.md
@@ -10,8 +10,8 @@ Inputs:
 |---|---|---|---|
 | `/holosoma/smplh_command` | `CmdSMPLH.msg` | SMPL-H 24-joint pose targets, retargeted to dense before the policy | 50Hz |
 | `/holosoma/dense_tracking_command` | `CmdDense.msg` | Dense per-joint 29-DoF target (q/dq + root quat) consumed directly by the policy | 50Hz |
-| `/cmd_vel` | `geometry_msgs/Twist` | Velocity command for the locomotion policy | any (timeout-guarded) |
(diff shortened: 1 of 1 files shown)


asetapen (approved):
Thank you for the updates!

src/holosoma_inference/holosoma_inference_service/README.md
```diff
@@ -3,41 +3,55 @@
...
 ## Service API
-Inputs (one of):
+
+Inputs:
 
 | Topic | Type | Description | Exp. Rate |
 |---|---|---|---|
 | `/holosoma/smplh_command` | `CmdSMPLH.msg` | SMPL-H 24-joint pose targets, retargeted to dense before the policy | 50Hz |
 | `/holosoma/dense_tracking_command` | `CmdDense.msg` | Dense per-joint 29-DoF target (q/dq + root quat) consumed directly by the policy | 50Hz |
+| `/cmd_vel` | `geometry_msgs/TwistStamped` | Velocity command for the locomotion policy | any (timeout-guarded) |
```
Kedaro, line 13: Not in the scope of this CR, but we should really consider prefixing `/cmd_vel` with `/holosoma` for consistency + to avoid name clash

Kedaro (approved):

src/holosoma_inference/holosoma_inference_service/README.md
kingb: Good point. The idea was many things already publish to /cmd_vel so holosoma would pick them up naturally... but it could be a re-mapping of /cmd_vel to /holosoma/cmd_vel to make it explicit that's what is desired. It's worth considering

PR #155, opened 2026-07-09
Cross-backend object physics domain randomization

juan-g-bonilla:
# PR 04 — Cross-backend object physics domain randomization

## Summary

Domain-randomize scene objects across all four backends, on top of the reproducible keyed sampler
(PR 02) and the scene-object registry (PR 03). Adds startup randomizers for object **mass**,
**material** (friction/restitution), **inertia**, and **freejoint damping** — each drawn through the
keyed `TermSampler`, so a `(term, env, episode)` draw is reproducible and per-env independent — plus
the typed per-backend config surface to drive them. Re-wires object DR into the shipped `wbt/g1`
`w_object` preset, which PR 03 had temporarily stripped.

The bulk of the PR is genuinely additive. **But it is NOT purely additive** — it also unifies the
granularity of the pre-existing *robot* friction term, which changes dynamics and per-seed results
for the `loco/g1` and `loco/t1` presets (see the callout below).

---

## ⚠️ Not result-preserving — `loco/g1` and `loco/t1` need re-baselining

**Robot friction DR (`randomize_friction_startup`) changed granularity on IsaacSim and MuJoCo**
— this is a *robot* term, unrelated to objects, that rides along in this PR.

- `terms/locomotion.py:967` now passes `per_env=True` (IsaacSim) and `terms/locomotion.py:985` now
  passes `shared_across_entities=True` (MuJoCo) into the friction draw. Before (03 base): the
  IsaacSim branch drew friction **per shape** and the MuJoCo branch **per geom**, independently.
  After: **one friction value per env**, broadcast across all of the robot's shapes/geoms.
- Impact: (1) realized contact dynamics differ — every robot geom now shares one coefficient instead
  of varying per link; (2) per-seed values differ even for shape 0, because the sampler draw drops
  the shape coordinate (`coords=(k,)` vs `(k, shape_ids)`; bucket pick `coords=(3,)` vs
  `(3, shape_ids)`).
- **Who is hit:** the shipped, enabled presets `loco/g1` (`friction_range=[0.5,1.25]`) and `loco/t1`
  (`friction_range=[0.1,1.0]`). Both preset files are byte-identical 03→04; the change is entirely in
  the term body. IsaacGym was already per-env (from PR 02) and is unchanged. There is **no opt-out**.
- Not masked: this is a `setup_terms` (startup) term, robot foot friction is never overwritten at
  reset. **Re-baseline `loco/g1` and `loco/t1`; do not expect a seed to reproduce a 03 run.**

Everything else in the PR is additive (new capability or new config surface); no *other* existing
preset's dynamics or reproducibility changes.

---

### Different dynamics (existing tasks)

- **Robot friction granularity** — see the callout. `terms/locomotion.py:967,985`;
  presets `loco/g1`, `loco/t1`.
- **`wbt/g1` `w_object` now randomizes the object again.** `object_state_dr_at_setup`
  (`config_values/wbt/g1/randomization.py:39`) is merged back into
  `g1_29dof_wbt_randomization_w_object.setup_terms` (`:141`). The 03 base had **no** object DR on
  this preset (setup was base-only). So the existing `g1_29dof_wbt_w_object` and
  `g1_29dof_wbt_fast_sac_w_object` experiments now randomize the spawned box's mass (`+[1.0,4.0]`),
  material friction (`[0.1,0.6]`), and inertia (`Ixx` scale `[0.5,1.5]`) at startup where 03 did not.
  On **IsaacSim** these ranges and the draw path are identical to pre-stack main (mass/inertia
  writers are byte-identical; `num_buckets=64`), so this is a *restoration* vs main — but still a
  dynamics change vs the 03 base. Not masked: the WBT command term overwrites only object pose/velocity
  from motion data, never mass/friction/inertia.

### Newly supported

- **Object mass / material / inertia DR is now cross-backend** (`terms/objects.py:157, 316, 502`).
  Each term has three real backend branches — IsaacGym (native
  `get/set_actor_rigid_{shape,body}_properties`), IsaacSim (project `events.py` writers), MuJoCo
  Warp+Classic (`randomize_field`, plus `scale_inertia_by_mass_ratio` for mass) — and raises
  `RandomizerNotSupportedError` on anything else. On pre-stack main these lived in `terms.locomotion`
  and were **IsaacSim-only** (raised on every other backend). Targets resolve via
  `_resolve_object_names` → registry `INDIVIDUAL` free bodies, defaulting to all of them, so a term
  no-ops cleanly on a robot-only scene.
- **Object freejoint-damping DR** (`randomize_object_{linear,angular}_damping_startup`,
  `terms/objects.py:757, 796`) — new terms supporting MuJoCo (Classic+Warp, via freejoint
  `dof_damping`) and IsaacSim (live-USD `PhysxRigidBodyAPI.{linear,angular}Damping`). See "latent"
  below — **no shipped preset wires them yet.**
- **MuJoCo mass DR now rescales object inertia by the mass ratio** and clamps mass to `1e-6`,
  matching the Isaac paths (`terms/objects.py` MuJoCo branch → `scale_inertia_by_mass_ratio`,
  `clamp(min=1e-6)`). This only reaches a user via the newly-wired `w_object` object DR above.

### Removed / fail-loud

- **Object damping on IsaacGym raises `RandomizerNotSupportedError`** (`terms/objects.py:727`,
  `_DAMPING_ISAACGYM_MSG`): IsaacGym has no runtime free-body-damping path (damping is import-time on
  `AssetOptions` only). Latent today (no preset wires damping), so no shipped IsaacGym run breaks; but
  a user who wires object damping and runs IsaacGym gets a hard, explanatory error. The message
  directs them to set damping via `PhysXPhysicsConfig` at spawn.

---

## Migration guide (breaking config/API surface — object-DR authors only)

Config-driven users who don't randomize objects and don't subclass DR terms see **no** API break
here (the friction-granularity change needs no config edit). For object-DR authors:

**1. Object-DR func paths and the inertia param type.** (Only relevant if you wired object DR against
*pre-stack main*; the 03 base had no object DR to migrate from.)

```python
# BEFORE (pre-stack main) — IsaacSim-only, in terms.locomotion
func="...terms.locomotion:randomize_object_rigid_body_mass_startup"
params={"mass_distribution_params": [1.0, 4.0]}
func="...terms.locomotion:randomize_object_rigid_body_inertia_startup"
params={"inertia_distribution_params_dict": {"Ixx": [0.5, 1.5]}}   # plain dict

# AFTER — cross-backend, in terms.objects
func="...terms.objects:randomize_object_rigid_body_mass_startup"
params={"mass_distribution_params": [1.0, 4.0]}
func="...terms.objects:randomize_object_rigid_body_inertia_startup"
params={"inertia_distribution_params_dict": InertiaScale(Ixx=[0.5, 1.5])}   # typed
```
`InertiaScale` / `MaterialRandomizationConfig` import from `holosoma.config_types.randomization`.
**Not a hard break:** `_coerce_inertia_params` still accepts a plain `{Ixx: ...}` dict and
`_coerce_material_cfg` still accepts a nested dict, so the typed classes are the preferred-but-optional
surface.

**2. Material DR is a new term with a typed, per-backend config** (no prior form). Each backend
sub-block names only the channels that backend honors — no silent cross-backend drop; MuJoCo has no
scalar restitution (it exposes `solref`/`solimp` instead):

```python
func="...terms.objects:randomize_object_rigid_body_material_startup"
params={"material": MaterialRandomizationConfig(
    isaacgym=IsaacGymMaterialDR(friction=[0.1, 0.6], restitution=[0.0, 1.0]),
    isaacsim=IsaacSimMaterialDR(static_friction=[0.1, 0.6], dynamic_friction=[0.1, 0.6], restitution=[0.0, 1.0]),
    mujoco=MujocoMaterialDR(sliding_friction=[0.1, 0.6]))}
```

**3. Internal writer signatures gained backward-compatible params (non-breaking).**
`isaacsim/events.py:randomize_rigid_body_material` gained `per_env: bool = False`;
`mujoco/backends/randomization.py:randomize_field` gained `shared_across_entities: bool = False`.
Both defaults preserve the old per-shape/per-entity behavior; only callers passing `True` (the robot
friction term, and the new object terms) change results. Custom callers need no change.

---

### DIFF
--- a/src/holosoma/holosoma/config_values/wbt/g1/randomization.py
+++ b/src/holosoma/holosoma/config_values/wbt/g1/randomization.py
@@ -2,6 +2,11 @@
 
 from holosoma.config_types.randomization import (
     BaseComRange,
+    InertiaScale,
+    IsaacGymMaterialDR,
+    IsaacSimMaterialDR,
+    MaterialRandomizationConfig,
+    MujocoMaterialDR,
     RandomizationManagerCfg,
     RandomizationTermCfg,
 )
@@ -27,6 +32,37 @@ robot_state_dr_at_setup = {
         params={
             "dof_pos_bias_range": [-0.01, 0.01],
             "enabled": True,
+        },
+    ),
+}
+
+object_state_dr_at_setup = {
+    "randomize_object_rigid_body_material_startup": RandomizationTermCfg(
+        func="holosoma.managers.randomization.terms.objects:randomize_object_rigid_body_material_startup",
+        params={
+            # Per-backend channels via the typed config — each backend names ONLY what it honors
+            # (no silent cross-backend drop). MuJoCo has no restitution channel (it's solref).
+            "material": MaterialRandomizationConfig(
+                isaacgym=IsaacGymMaterialDR(friction=[0.1, 0.6], restitution=[0.0, 1.0]),
+                isaacsim=IsaacSimMaterialDR(
+                    static_friction=[0.1, 0.6], dynamic_friction=[0.1, 0.6], restitution=[0.0, 1.0]
+                ),
+                mujoco=MujocoMaterialDR(sliding_friction=[0.1, 0.6]),
+            ),
+        },
+    ),
+    "randomize_object_rigid_body_mass_startup": RandomizationTermCfg(
+        func="holosoma.managers.randomization.terms.objects:randomize_object_rigid_body_mass_startup",
+        params={
+            "mass_distribution_params": [1.0, 4.0],
+        },
+    ),
+    "randomize_object_rigid_body_inertia_startup": RandomizationTermCfg(
+        func="holosoma.managers.randomization.terms.objects:randomize_object_rigid_body_inertia_startup",
+        params={
+            # Only Ixx is randomized, reproducing beyondmimic. Unnamed components default to
+            # identity [1.0, 1.0]; name more to randomize them.
+            "inertia_distribution_params_dict": InertiaScale(Ixx=[0.5, 1.5]),
         },
     ),
 }
@@ -99,11 +135,10 @@ g1_29dof_wbt_randomization = RandomizationManagerCfg(
     step_terms={**base_step_terms},
 )
 
-# Same as the base preset on this branch — the object-DR follow-up branch adds the
-# object_state_dr_at_setup terms (material/mass/inertia) into this preset's setup_terms.
 g1_29dof_wbt_randomization_w_object = RandomizationManagerCfg(
     setup_terms={
         **base_setup_terms,
+        **object_state_dr_at_setup,
     },
     reset_terms={
         **base_reset_terms,
--- a/src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
+++ b/src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
@@ -196,6 +196,7 @@ def randomize_field(
     entity_names: list[str] | None = None,
     entity_type: str | None = None,
     operation: Literal["add", "scale", "abs"] = "abs",
+    shared_across_entities: bool = False,
     axis_base: int = 0,
     num_buckets: int | None = None,
 ) -> None:
@@ -236,6 +237,11 @@ def randomize_field(
         Required if entity_names is provided, otherwise inferred from field
     operation : Literal["add", "scale", "abs"]
         Operation to apply: add to current, scale current, or set absolute value
+    shared_across_entities : bool
+        If True, every entity gets the SAME per-env sample (the first entity's), instead of an
+        independent draw each. Use to randomize a group of entities as one logical unit — e.g.
+        a freejoint's 3 linear ``dof_damping`` DOFs sharing one value (PhysX exposes a single
+        linear damping scalar, so this keeps the concept aligned across backends).
     num_buckets : Optional[int]
         Quantization knob mirroring the PhysX backends' material bucketing, for a config that wants
         the SAME staircase discretization on MuJoCo. ``None`` (default): draw continuously — MuJoCo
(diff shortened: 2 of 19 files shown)


src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
```diff
@@ -400,6 +410,11 @@ def randomize_field(
             per_axis.append(column[bucket_ids])
     random_values = torch.stack(per_axis, dim=-1)  # (n_e, n_n, n_a)
 
+    # Share one per-env sample across the whole entity group (the first entity's draw), so a
+    # multi-DOF/entity unit is randomized as one value rather than independently per entity.
```
samuelgundry, line 414: > multi-DOF/entity unit is randomized as one value rather than independently per entity.

Naively, what's the benefits of this?  Just fewer variables/simpler OR is this something more fundamental?

src/holosoma/holosoma/config_values/wbt/g1/randomization.py
```diff
@@ -31,6 +36,37 @@
...
+                mujoco=MujocoMaterialDR(sliding_friction=[0.1, 0.6]),
+            ),
+        },
+    ),
+    "randomize_object_rigid_body_mass_startup": RandomizationTermCfg(
+        func="holosoma.managers.randomization.terms.objects:randomize_object_rigid_body_mass_startup",
+        params={
+            "mass_distribution_params": [1.0, 4.0],
+        },
+    ),
```
samuelgundry, line 59: I don't have a good sense how we'll configure per-object or per-object-kind/path DRs.  Which is fine for now, and I suspect could be quite cumbersome with the existing framework. Maybe not.

samuelgundry (approved):
Looks good. I have a naive Q below w.r.t per-entity vs per-link randomized values.

src/holosoma/holosoma/simulator/mujoco/backends/randomization.py
juan-g-bonilla: So, before this PR, I believe one backend did all the robot and the others did per-geometry/per-body randomization. I chose one consistent behavior that I felt was reasonable. I think the more intuitive behavior with this DR is "I'm not sure how grippy my feet are" vs. "I think my two feet might have different grippy-ness". That being said I think it should be easy to add the "all geoms different properties" DR term.

src/holosoma/holosoma/config_values/wbt/g1/randomization.py
juan-g-bonilla: Right now, you can pass:
```
object_names = ["foo", "bar"]
```
to these terms, which means randomize all these bodies (with different RNGs). If you don't pass anything, all bodies are randomized.

PR #156, opened 2026-07-09
feat: terrain hide_visual, frequency-string rate model, backend-aware run_sim device

juan-g-bonilla:
## Summary

Three leaf features pulled out of the mega scene-objects branch into their own PR:
1. **terrain `hide_visual`** — render terrain invisible while keeping its collider (additive, opt-in config field).
2. **frequency-string rate model** — `control_decimation` / `render_interval` accept an int OR a target frequency string (`"50Hz"`, `">50Hz"`, `"<50Hz"`) resolved against `fps`; callers read the resolved integer via new `control_decimation_steps` / `render_interval_steps` properties. Result-preserving for every shipped preset.
3. **run_sim backend-aware default device** — `RunSimConfig.device` default `"cpu"` → `None`, resolved per backend. **This changes the resolved device for IsaacSim/IsaacGym.**

## ⚠️ Not result-preserving — re-baseline affected runs

**`run_sim` on IsaacSim/IsaacGym with no `--device`, on a CUDA-capable box, now runs on `cuda:0` where it previously ran on `cpu`.** The old logic only upgraded the `"cpu"` default to `cuda:0` for the MuJoCo-Warp backend; IsaacSim/IsaacGym (which default `mujoco_backend=CLASSIC`) stayed on `cpu`. The new logic resolves `device=None` to `cuda:0` for those backends when `torch.cuda.is_available()`. GPU vs CPU PhysX are different solver pipelines, so dynamics/repro differ for anyone who relied on the old unset-`--device` default. Trigger is narrow: `run_sim` entrypoint only, GPU box, no explicit `--device`. `run_sim.py:35-64`, `run_sim.py:71` → `sim_utils.py:262-263` → `base_simulator.py:172` → `isaacgym.py:128-139` / `isaacsim.py:102-105`.

### Different dynamics / reproducibility
- **IsaacSim/IsaacGym default device `cpu` → `cuda:0`** (GPU box, unset `--device`, `run_sim`). See callout above. `config_types/run_sim.py:88`, `run_sim.py:35-64`.

### Newly supported (opt-in; no-op default — no existing task/env affected)
- **terrain `hide_visual: bool = False`** — hides terrain visual while keeping its collider. No path touches collision (`contype`/`conaffinity`/colliders untouched). `config_types/terrain.py` (default `False`), `managers/terrain/base.py` (property).
  - **MuJoCo:** honored for **all** mesh types — zero-alpha `geom.rgba=[0,0,0,0]` in the shared post-dispatch block. `simulator/mujoco/scene_manager.py:219-221`.
  - **IsaacSim:** honored for **`plane` terrain only** — `_hide_prim_subtree` authors `visibility=invisible` on the `/World/ground` subtree. **`trimesh`/`load_obj` terrain silently ignores `hide_visual`** (the mesh branch never calls it — no warning, no effect). `simulator/isaacsim/isaacsim.py:73-86, 457-460`.
  - **IsaacGym:** **never actually hides.** Headless is a trivial no-op (nothing drawn); headful logs an "unsupported" warning and leaves the ground visible. `simulator/isaacgym/isaacgym.py:201-210`.

### Removed / fail-loud (config surface)
- **`control_decimation` / `render_interval` retyped `int` → `DecimationLike` (`int | str`)** with a new `SimEngineConfig` `model_validator` (`_validate_rates`). The base had **no** validation on these fields; now a malformed rate raises `pydantic.ValidationError` **at config construction**: `int < 1` (e.g. `0`, `-3` — previously accepted) and unparseable / non-exact-divisor strings (`"banana"`, bare `"60Hz"` at `fps=200`). `config_types/simulator.py:123,143,152-160`, `config_types/frequency.py`.
- **Resolved-int read path moved to `*_steps` properties.** Any code that read the raw `control_decimation` / `render_interval` as an int must switch to `control_decimation_steps` / `render_interval_steps` (the field now holds the value as written, which may be a string). All in-tree callers already migrated. `config_types/simulator.py:163-172`.

### Latent edge-case (no shipped preset triggers it)
- **`run_sim` explicit `--device cpu` + mjwarp:** old logic force-upgraded any `"cpu"` to `cuda:0` for WARP; new logic honors an explicit `cpu`. Only reachable by explicitly passing `--device cpu` to a Warp run, which is not a valid config (Warp is CUDA-only). `run_sim.py:35-47`.

_Note: the IsaacSim `link_physics.isaacsim` NameError fix that previously rode along in this PR's imports has been moved to its origin, **PR 03** (`fix(isaacsim): import stage_utils and Usd for link_physics material bind`) — that's where `_bind_robot_link_material` and its unimported use were introduced. PR 05 now only adds `UsdGeom` (needed by its own `_hide_prim_subtree`); `stage_utils`/`Usd` arrive via PR 03. This PR's final tree is unchanged by the move._

## Migration guide (API/config surface)

**Rate fields — read the resolved `*_steps` property**
```python
# BEFORE                                      # AFTER
for _ in range(sim.control_decimation):       for _ in range(sim.control_decimation_steps):
sim_fps / sim.control_decimation              sim_fps / sim.control_decimation_steps
render_interval < control_decimation          render_interval_steps < control_decimation_steps
```
Fields now accept a frequency string resolved against `fps`: `"50Hz"` requires an exact decimation (errors if `fps/target` isn't whole), `">50Hz"` floors (achieved rate ≥ target), `"<50Hz"` ceils (achieved rate ≤ target). Ints ≥ 1 pass through unchanged. `int < 1` and malformed strings now raise `ValidationError` at construction.

**run_sim device default (currently undocumented in the migration guide — add it)**
```python
# BEFORE                                      # AFTER
RunSimConfig.device: str | None = "cpu"       RunSimConfig.device: str | None = None
# unset --device:                             # unset --device resolves per backend:
#   mjwarp            -> cuda:0                #   mjwarp             -> cuda:0        (unchanged)
#   isaacsim/isaacgym -> cpu                   #   isaacsim/isaacgym  -> cuda:0 if CUDA else cpu  (CHANGED)
#   classic mujoco    -> cpu                   #   classic mujoco     -> cpu          (unchanged)
```
Pass an explicit `--device cpu` (or `--device cuda:1`) to pin the old behavior. IsaacSim/IsaacGym users who want the previous CPU default **must** now pass `--device cpu`.

**terrain `hide_visual`** — additive field, default `False`; no migration required. If you set it `True`, note the per-backend support matrix above (MuJoCo: all; IsaacSim: plane only; IsaacGym: unsupported/warns).

### DIFF
--- a/src/holosoma/holosoma/config_types/frequency.py
+++ b/src/holosoma/holosoma/config_types/frequency.py
@@ -1,0 +1,97 @@
+"""Resolve rate fields written as frequency strings ("20Hz") into decimations.
+
+A decimation is "act every Nth tick of a base clock" running at ``base_hz``. A rate field is either a
+decimation int (every Nth tick) or a target frequency string resolved against a base rate.
+
+Grammar (case-insensitive, surrounding whitespace ignored)::
+
+    "20Hz"     require an exact decimation (base/20 must be a whole number); error otherwise
+    ">20Hz"    round to achieve a rate >= target (floor(base/20))
+    "<20Hz"    round to achieve a rate <= target (ceil(base/20))
+"""
+
+from __future__ import annotations
+
+import math
+import re
+from typing import Union
+
+from loguru import logger
+
+DecimationLike = Union[int, str]
+"""A rate field: a decimation int (every Nth base tick), or a frequency string resolved against a base
+rate via :func:`resolve_decimation`. Form-checked by :func:`validate_decimation_like`."""
+
+# an optional comparison prefix (< or >), a positive number (int or float), then a Hz unit.
+_FREQ_RE = re.compile(r"^\s*([<>]?)\s*([0-9]*\.?[0-9]+)\s*[Hh][Zz]\s*$")
+
+
+def is_frequency_string(value: DecimationLike) -> bool:
+    """Whether ``value`` is a frequency string (vs an already-numeric decimation)."""
+    return isinstance(value, str)
+
+
+def validate_decimation_like(value: DecimationLike, *, field: str) -> None:
+    """Validate form only (int >= 1, or a well-formed freq string), without a base rate.
+
+    Resolution to an int happens later against the base rate. Raises ValueError on a bad form.
+    """
+    if isinstance(value, bool):  # bool is an int subclass; reject explicitly
+        raise ValueError(f"{field}: expected a decimation int or frequency string, got bool {value!r}.")
+    if isinstance(value, int):
+        if value < 1:
+            raise ValueError(f"{field}: decimation must be >= 1, got {value}.")
+        return
+    if not isinstance(value, str) or _FREQ_RE.match(value) is None:
+        raise ValueError(f"{field}: invalid {value!r}; expected an int >= 1 or e.g. '20Hz', '>20Hz', '<20Hz'.")
+
+
+def resolve_decimation(value: DecimationLike, base_hz: float, *, field: str, log: bool = False) -> int:
+    """Resolve ``value`` to an integer decimation >= 1 against a ``base_hz`` clock.
+
+    An int passes through unchanged (>= 1 enforced). A bare frequency string ("20Hz") requires an
+    exact decimation (``base/target`` must be a whole number) and raises otherwise. A ">"-prefixed
+    string (">20Hz") floors ``base/target`` so the achieved rate is >= target; a "<"-prefixed string
+    ("<20Hz") ceils it so the achieved rate is <= target. A target faster than ``base_hz`` clamps to
+    1 (warned when ``log``). ``log`` is passed once at config validation; resolved-property reads stay
+    silent (they re-resolve on every access, which would spam per-step logs).
+    """
+    if isinstance(value, bool):  # bool is an int subclass; reject explicitly
+        raise ValueError(f"{field}: expected a decimation int or frequency string, got bool {value!r}.")
+    if isinstance(value, int):
+        if value < 1:
+            raise ValueError(f"{field}: decimation must be >= 1, got {value}.")
+        return value
+    if not isinstance(value, str):
+        raise ValueError(f"{field}: expected a decimation int or frequency string, got {value!r}.")
+
+    m = _FREQ_RE.match(value)
+    if m is None:
+        raise ValueError(f"{field}: invalid frequency {value!r}; expected e.g. '20Hz', '>20Hz', '<20Hz'.")
+    prefix = m.group(1)
+    hz = float(m.group(2))
+    if hz <= 0.0:
+        raise ValueError(f"{field}: frequency must be positive, got {value!r}.")
+
+    ratio = base_hz / hz
+    if prefix == ">":
+        dec = math.floor(ratio)  # floor so achieved rate >= target
+    elif prefix == "<":
+        dec = math.ceil(ratio)  # ceil so achieved rate <= target
+    else:
+        # Bare "20Hz": require an exact decimation, else the achieved rate would silently differ.
+        dec = round(ratio)
+        if not math.isclose(ratio, dec, rel_tol=1e-9, abs_tol=1e-9):
+            raise ValueError(
+                f"{field}: {value} is not exactly achievable from base {base_hz:g}Hz "
+                f"(base/target = {ratio:g}). Use '>{hz:g}Hz' (achieve >= target) or "
+                f"'<{hz:g}Hz' (achieve <= target) to allow rounding."
+            )
+
+    if dec < 1:
+        if log:
+            logger.warning(f"{field}: requested {value} but base clock is {base_hz:g}Hz; capping at decimation=1.")
+        dec = 1
+    if log:
+        logger.info(f"{field}: {value} @ base {base_hz:g}Hz -> decimation={dec} (achieved {base_hz / dec:g}Hz).")
+    return dec
--- a/src/holosoma/holosoma/config_types/run_sim.py
+++ b/src/holosoma/holosoma/config_types/run_sim.py
@@ -85,6 +85,8 @@ class RunSimConfig:
     Only used by run_sim.py for real-time display synchronization.
     """
 
-    device: str | None = "cpu"
-    """Device to use for simulation. None auto-detects based on the simulator type.
+    device: str | None = None
+    """Device to use for simulation. None (the default) auto-detects based on the simulator
+    backend: cuda:0 for MuJoCo Warp, cuda:0 for IsaacSim/IsaacGym when CUDA is available (else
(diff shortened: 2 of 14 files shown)


src/holosoma/holosoma/config_types/frequency.py
```diff
@@ -0,0 +1,97 @@
...
+
+def resolve_decimation(value: DecimationLike, base_hz: float, *, field: str, log: bool = False) -> int:
+    """Resolve ``value`` to an integer decimation >= 1 against a ``base_hz`` clock.
+
+    An int passes through unchanged (>= 1 enforced). A bare frequency string ("20Hz") requires an
+    exact decimation (``base/target`` must be a whole number) and raises otherwise. A ">"-prefixed
+    string (">20Hz") floors ``base/target`` so the achieved rate is >= target; a "<"-prefixed string
+    ("<20Hz") ceils it so the achieved rate is <= target. A target faster than ``base_hz`` clamps to
+    1 (warned when ``log``). ``log`` is passed once at config validation; resolved-property reads stay
+    silent (they re-resolve on every access, which would spam per-step logs).
```
samuelgundry, line 57: Curious, did a use case pop-up for this?   It's nice but we should be cautious where adding features just for features sake (especially with agent-based coding now, can easily create a lot of code/entropy/stuff...)

src/holosoma/holosoma/config_types/run_sim.py
```diff
@@ -85,6 +85,8 @@ class RunSimConfig:
     Only used by run_sim.py for real-time display synchronization.
     """
 
-    device: str | None = "cpu"
-    """Device to use for simulation. None auto-detects based on the simulator type.
+    device: str | None = None
+    """Device to use for simulation. None (the default) auto-detects based on the simulator
+    backend: cuda:0 for MuJoCo Warp, cuda:0 for IsaacSim/IsaacGym when CUDA is available (else
+    cpu), and cpu for classic MuJoCo. Pass an explicit value (e.g. "cpu" or "cuda:1") to override.
     """
```
samuelgundry, line 92: Good one. We added auto-detect for mjwarp just before public release and didn't backfill to gym/lab.

src/holosoma/holosoma/simulator/isaacgym/isaacgym.py
```diff
@@ -198,11 +198,20 @@ def get_supported_scene_formats(self) -> list[str]:
...
+        terrain_state = self.terrain_manager.get_state("locomotion_terrain")
+        mesh_type = terrain_state.mesh_type
+        # IsaacGym's ground has no separable visual, so hide_visual is satisfied only where it is
+        # already met (headless draws nothing) and warned otherwise. set_headless runs before this,
+        # so self.headless is reliable.
+        if terrain_state.hide_visual and not self.headless:
+            logger.warning(
+                "terrain hide_visual=True is not supported on IsaacGym in headful mode (its ground "
+                "geometry has no separable visual); run headless or use MuJoCo/IsaacSim."
+            )
```
samuelgundry, line 210: What's the use case for hiding terrain?  Is this a work-around where terrain is "required" and we want to load scenes? 

Edit, I see the comment below.  Why do we want to keep the terrain/ground collider?
```
# Hide the ground plane's visual while keeping its collider; avoids z-fighting a scene
# USD's own floor. Applied to the whole /World/ground subtree.
```

src/holosoma/holosoma/run_sim.py
```diff
@@ -30,17 +30,36 @@ def run_simulation(config: RunSimConfig):
...
+            if torch.cuda.is_available():
+                resolved_device = "cuda:0"
+                logger.info(f"Auto-detected {sim_name} backend with CUDA available - setting device to cuda:0")
+            else:
+                resolved_device = "cpu"
+                logger.warning(f"{sim_name} backend selected but CUDA is unavailable - defaulting to cpu")
+        else:
+            # Classic MuJoCo (CPU physics engine).
+            resolved_device = "cpu"
+            logger.info(f"Auto-detected {sim_name} backend - setting device to cpu")
```
samuelgundry, line 60: I feel like this resolution should be inside each backend itself -- decoupled, easier to check/catch when adding new simulators w.r.t a contract eg. "backend.resolve_device()"

samuelgundry (approved):
Nothing blocking, a minor comment w.r.t device resolution location so approved. Perhaps add explanatory reasons WHY the terrain collider is kept but visual is hiddden -- it's not super obvious if this is a hack/work-around in lieu of a bigger idealistic refactor.

src/holosoma/holosoma/config_types/frequency.py
juan-g-bonilla: this is mostly in preparation of the sensor and telemetry work coming downstream. Those models have a specific rate at which we will want them to run/publish, e.g. 50Hz. If we configure them to achieve that rate through a decimation factor, a) it's confusing (you have to do the mental math that if physics fps=500Hz, and control decimation is 5, then you have to set sensor decimation to 2 to achieve 50Hz) and b) if we end up changing the physics fps because we want to improve performance by trading off some integration error, that triggers a cascade of every single thing downstream effectively changing the rate at which things run. Setting these Hz manually is clearer for users AND prevents one config change cascading into a lot of changed behavior.

In this case, I was aware of this behavior and actually went back and forth on it a bit. I did add the < and > behavior a bit defensively, not sure if we'll need it.

src/holosoma/holosoma/run_sim.py
juan-g-bonilla: Yep, good point. Just kept it where the mjwarp resolution was happening, but the scalable thing is to make part of the `base_simulator.py` contract.

juan-g-bonilla: I'll fix it in a follow up branch so we can get this out without requiring a re-stamp from you.

src/holosoma/holosoma/simulator/isaacgym/isaacgym.py
juan-g-bonilla: The best approach, if you scene has a floor with collision, is to leave the terrain completely off and then just do collision on the scene assets. However, not all scenes will have collision setup for their floor. For example, the residential scenes we have in our SDG browser have no collision configured on the floor. Also, 3DGS assets will generally not have collision meshes, which is generally fine if we want them to be floors and backgrounds, but they do need a extra ground collision. One could argue that the proper solution then is to add a collision mesh to these assets, but out of convenience...

src/holosoma/holosoma/run_sim.py
juan-g-bonilla: Actually, I walk that back. This is tricky because we need to know the device before we initialize the isaac simulation app, which we need to do before we can import and use the isaac simulator class. So, we can't really include the device resolution logic in the simulator classes because those are not importable before we know the device. Pushing this for another time.

PR #158, opened 2026-07-10
Hook system for simulation callbacks

juan-g-bonilla:
## Summary

Replaces the hard-coded step-loop side effects scattered across the simulator backends with a single lifecycle **hook registry**. Instead of each backend hand-calling `virtual_gantry.step()`, `_step_bridge()`, and `capture_video_frame()` inside its `simulate_at_each_physics_step`, and overriding `on_episode_start/end`, the runtime participants now register callbacks against typed `Phase` events that `BaseTask` (and the `run_sim` `DirectSimulation` loop) emit.

1. **`hooks.py` — new `HookRegistry` + `Phase` enum.** Seven lifecycle phases: `FRAME_BEGIN` / `FRAME_END` bracket the outer control-rate tick, `PRE_STEP` / `POST_STEP` bracket each physics substep, plus `EPISODE_START`, `EPISODE_END`, `CLOSE`. Deterministic registration-order dispatch, per-phase payload-arity validation, `enable/disable/remove` handles, mutation-during-emit guards, and a reverse-order `CLOSE` that runs every hook exactly once and aggregates failures into one `HookCloseError`. Covered by `tests/test_hooks.py` (`no_sim`).
2. **Per-hook cadence — `every`.** `hooks.add(phase, fn, every=4)` runs a callback once per N emissions of its phase. On periodic phases (`FRAME_*`, `*_STEP`), `every` also accepts a frequency string (`"30Hz"`, `">30Hz"`, `"<30Hz"`) resolved against the phase's base tick rate, which `BaseSimulator._hook_base_rates()` supplies from the sim config: `fps` for the substep phases, `fps / control_decimation_steps` for the frame phases. `HookHandle.set_every()` retunes a live hook (counter resets). Event phases (episode) accept int decimation only; `CLOSE` rejects any decimation.
3. **Participants self-register.** `SimulatorBridge`, `VirtualGantry`, and the video recorder each gain a `register_hooks()` that wires their existing methods to phases. The backends call `participant.register_hooks(self.hooks)` at setup; their step methods lose the inline calls.
4. **Deterministic teardown.** New `BaseSimulator.close()` (idempotent) emits `Phase.CLOSE`; `train_agent.py` closes the env in a `finally`, and `DirectSimulation.cleanup()` routes through `env.close()` / `simulator.close()`. Adds `SimulatorBridge.close()` (closes the clock-publisher socket) — **it did not previously exist**, and registering it for `CLOSE` is what surfaced the bug below.

Behavior-preserving for every shipped run: the same callbacks fire at the same rate, in the same order.

### Newly present (opt-in / internal; no shipped run affected)
- **`Phase.CLOSE` teardown path.** `BaseSimulator.close()` now emits `CLOSE`, running `bridge.close` (clock socket) and `video.cleanup` in reverse registration order, idempotently. Previously teardown was ad-hoc: `DirectSimulation.cleanup` only called `video_recorder.cleanup()`, and the bridge's clock socket was never explicitly closed. Both `train_agent.py` (new `finally`) and `DirectSimulation` now go through `close()`.
- **`HookRegistry` public surface on `BaseSimulator.hooks`.** New extension point — downstream code can register per-phase callbacks (with an optional cadence) without subclassing a backend or overriding `on_episode_*`.
- **`run_sim` loop emits the same phases as training.** `DirectSimulation.run()` emits `PRE_STEP`/`POST_STEP` every physics step and `FRAME_BEGIN`/`FRAME_END` every `control_decimation_steps`, so hooks behave identically under `run_sim` and `BaseTask`.

### Removed (internal API surface)
- **`BaseSimulator.on_episode_start` / `on_episode_end` / `capture_video_frame` / `_step_bridge` deleted.** These were the override seams for episode/video/bridge behavior. Callers now emit phases: `base_task.py` emits `EPISODE_START` / `EPISODE_END` (replacing the `hasattr(simulator, "on_episode_*")` guards) and the frame/step phases; `sim_utils.py` emits `EPISODE_START` at init. Any out-of-tree subclass that overrode `on_episode_start/end` to hook episode boundaries must migrate to `simulator.hooks.add(Phase.EPISODE_START, ...)`. All in-tree callers migrated.
- **Video recorder's manual decimation counter deleted.** `capture_frame` no longer tracks `_frame_counter % control_decimation`; it registers on `FRAME_END`, which already fires once per frame at control rate.

### Result-preserving (verified)
- **Per-step callback timing unchanged.** `bridge.step` and `virtual_gantry.step` register on `PRE_STEP` (fires every physics substep, before `simulate_at_each_physics_step`, as before — SDK torques and gantry forces must land before the substep integrates them). Registration order — gantry before bridge (gantry registered ahead of `_init_bridge` in every backend) — matches the old inline call order.
- **Video capture rate unchanged.** `video.capture_frame` registers on `FRAME_END`: once per frame, at control frequency — the same rate the old per-substep counter decimated to. The capture point moves from "after the Nth physics substep" to "after `_refresh_sim_tensors` at frame end".
- **Episode-start order unchanged.** Gantry's `set_position_to_robot` (now `VirtualGantry.on_episode_start`, `env_id == 0` guarded) registers before the video recorder's `on_episode_start`, matching the old `BaseSimulator.on_episode_start` body.

## Migration guide (out-of-tree only)

**Episode / video / bridge hooks — register instead of override**
```python
# BEFORE (subclass override)                 # AFTER (register a callback)
class MySim(MuJoCo):                          sim.hooks.add(Phase.EPISODE_START, my_fn)   # my_fn(env_id)
    def on_episode_start(self, env_id=0):     sim.hooks.add(Phase.POST_STEP, my_capture)
        super().on_episode_start(env_id)      sim.hooks.add(Phase.FRAME_END, my_log, every="10Hz")
        my_fn(env_id)                         # deregister later via the returned handle:
                                              handle = sim.hooks.add(Phase.CLOSE, my_cleanup); handle.remove()
```
`simulator.capture_video_frame()` / `_step_bridge()` are gone — the video recorder and bridge register themselves; there is nothing to call. Payload arity is validated at emit: `EPISODE_START`/`EPISODE_END` pass `env_id`; all other phases pass no args.

### DIFF
--- a/src/holosoma/holosoma/simulator/base_simulator/hooks.py
+++ b/src/holosoma/holosoma/simulator/base_simulator/hooks.py
@@ -1,0 +1,261 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Callable, Mapping
+
+from typing_extensions import Self
+
+from holosoma.config_types.frequency import DecimationLike, is_frequency_string, resolve_decimation
+
+
+class Phase(str, Enum):
+    """Lifecycle points emitted by the active simulator loop.
+
+    ``payload`` names the positional args ``emit`` forwards to callbacks. ``periodic`` marks phases
+    that fire on a fixed clock (per-substep or per-frame ticks) — only these accept a frequency-string
+    rate (``"30Hz"``) for :meth:`HookRegistry.add`'s ``every``; event phases (episode/close) accept an
+    int decimation ("every Nth event") but reject frequency strings, which have no clock to resolve
+    against.
+
+    ``FRAME_*`` bracket the outer tick (once per frame); ``*_STEP`` bracket each physics substep. The
+    names are engine-agnostic — they don't assume what drives the outer tick (a policy, a bridge, ...).
+    """
+
+    payload: tuple[str, ...]
+    periodic: bool
+
+    FRAME_BEGIN = ("frame_begin", (), True)
+    PRE_STEP = ("pre_step", (), True)
+    POST_STEP = ("post_step", (), True)
+    FRAME_END = ("frame_end", (), True)
+    EPISODE_END = ("episode_end", ("env_id",), False)
+    EPISODE_START = ("episode_start", ("env_id",), False)
+    CLOSE = ("close", (), False)
+
+    def __new__(cls, value: str, payload: tuple[str, ...], periodic: bool) -> Self:
+        obj = str.__new__(cls, value)
+        obj._value_ = value
+        obj.payload = payload
+        obj.periodic = periodic
+        return obj
+
+
+class HookRegistryError(RuntimeError):
+    """Raised when the hook registry is misused."""
+
+
+class HookCloseError(RuntimeError):
+    """Raised after close hooks run if one or more hooks failed."""
+
+    def __init__(self, failures: list[tuple[str, Exception]]) -> None:
+        self.failures = failures
+        details = ", ".join(f"{name}: {exc!r}" for name, exc in failures)
+        super().__init__(f"{len(failures)} close hook(s) failed: {details}")
+
+
+@dataclass
+class _HookRecord:
+    phase: Phase
+    callback: Callable[..., Any]
+    name: str
+    every: int = 1
+    """Resolved emission decimation: the callback runs once per ``every`` emissions of its phase."""
+    _counter: int = 0
+    """Emissions counted toward the next fire; only advances while enabled (disabled hooks leave the snapshot)."""
+    enabled: bool = True
+    removed: bool = False
+
+    def due(self) -> bool:
+        """Advance the per-hook counter and report whether this emission should fire the callback."""
+        self._counter += 1
+        if self._counter >= self.every:
+            self._counter = 0
+            return True
+        return False
+
+
+class HookHandle:
+    """Identity handle returned by hook registration."""
+
+    def __init__(self, registry: HookRegistry, record: _HookRecord) -> None:
+        self._registry = registry
+        self._record = record
+
+    @property
+    def name(self) -> str:
+        return self._record.name
+
+    @property
+    def phase(self) -> Phase:
+        return self._record.phase
+
+    @property
+    def enabled(self) -> bool:
+        return self._record.enabled and not self._record.removed
+
+    @property
+    def every(self) -> int:
+        """Current resolved emission decimation (see :meth:`set_every`)."""
+        return self._record.every
+
+    def enable(self) -> None:
+        self._registry._enable(self._record)
+
+    def disable(self) -> None:
+        self._registry._disable(self._record)
+
+    def set_every(self, every: DecimationLike) -> None:
+        """Change this hook's cadence live (int decimation or frequency string), resetting its counter.
+
+        Same rules as ``every`` at registration. Safe to call while other phases emit, but not from
+        inside this hook's own phase (mutating a phase mid-emit raises)."""
+        self._registry._set_every(self._record, every)
+
+    def remove(self) -> None:
+        self._registry._remove(self._record)
+
+
+class HookRegistry:
+    """Small lifecycle hook registry with deterministic registration order."""
+
+    def __init__(self, base_rates: Mapping[Phase, float] | None = None) -> None:
+        self._records: dict[Phase, list[_HookRecord]] = {phase: [] for phase in Phase}
+        self._snapshots: dict[Phase, tuple[_HookRecord, ...]] = dict.fromkeys(Phase, ())
+        self._emitting: set[Phase] = set()
+        self._closed = False
+        self._base_rates: dict[Phase, float] = dict(base_rates or {})
+
+    def add(
+        self,
+        phase: Phase,
+        callback: Callable[..., Any],
+        *,
+        name: str | None = None,
+        every: DecimationLike = 1,
+    ) -> HookHandle:
+        """Register a hook callback for a lifecycle phase.
+
+        ``every`` sub-samples emissions: an int decimation runs the callback once per ``every``
+        emissions; a frequency string (``"30Hz"``, ``">30Hz"``, ``"<30Hz"``) is resolved against the
+        phase's base tick rate into that decimation. Frequency strings require a periodic phase and a
+        known base rate; event phases (episode/close) accept only int decimations. Default ``1`` fires
+        every emission.
+        """
+        self._assert_mutable(phase)
+        record = _HookRecord(
+            phase=phase,
+            callback=callback,
+            name=name or self._default_name(callback),
+            every=self._resolve_every(phase, every, name),
+        )
+        self._records[phase].append(record)
+        self._rebuild(phase)
+        return HookHandle(self, record)
+
+    def _resolve_every(self, phase: Phase, every: DecimationLike, name: str | None) -> int:
+        """Turn a hook's ``every`` into an int decimation, resolving frequency strings vs the base rate."""
+        field = f"hook {name or '<callback>'!r} on phase {phase.value!r} 'every'"
+        if phase is Phase.CLOSE and every != 1:
+            raise HookRegistryError(f"{field}: CLOSE fires once at teardown; decimating it would skip cleanup.")
+        if is_frequency_string(every):
+            if not phase.periodic:
+                raise HookRegistryError(
+                    f"{field}: phase {phase.value!r} is not periodic; use an int decimation, not {every!r}."
+                )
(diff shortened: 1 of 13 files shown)


src/holosoma/holosoma/simulator/isaacgym/isaacgym.py
```diff
@@ -371,12 +371,14 @@ def create_envs(self, num_envs, env_origins, base_init_state):
...
             cfg=gantry_cfg,
         )
+        self.virtual_gantry.register_hooks(self.hooks)
 
         # Initialize bridge system using base class helper
         self._init_bridge()
 
         if self.video_recorder:
             self.video_recorder.setup_recording()
+            self.video_recorder.register_hooks(self.hooks)
```
samuelgundry, line 381: Probably over-engineering for _now_: even these `init_bridge()` and `video_recorder.setup_recording()` should arguably be abstracted into a hook system. Not a blocker, curious if this is a intuitive extension with your given framework, perhaps Setup() and Register(), akin/like UE OnConstruction() and BeginPlay() hooks. These are separate because the former is  "I can set myself up independent of the rest of the env/scene" and the latter is "Everyone is setup, references between components/entities/systems are valid".  Doing so means fewer custom code orchestration/ordering in the backend simulators.

src/holosoma/holosoma/simulator/base_simulator/hooks.py
```diff
@@ -0,0 +1,257 @@
...
+
+    payload: tuple[str, ...]
+    periodic: bool
+
+    CONTROL_PRE_PHYSICS = ("control_pre_physics", (), True)
+    PHYSICS_PRE_SIMULATE = ("physics_pre_simulate", (), True)
+    PHYSICS_POST_SIMULATE = ("physics_post_simulate", (), True)
+    CONTROL_POST_REFRESH = ("control_post_refresh", (), True)
+    EPISODE_END = ("episode_end", ("env_id",), False)
+    EPISODE_START = ("episode_start", ("env_id",), False)
```
samuelgundry, line 34: We may want to use batch `env_ids` for episode hooks?  Subscribers/consumers can then choose to iterate or not?   e.g, emit the tensor from base task, then gantry and video recorder iterate?

samuelgundry (changes requested):
This is such a good cleanup. As discussed offline, I strongly prefer a hook system akin to game engines -- it's a proven design pattern.  I haven't approved as I think we want to use tensor/batch env_ids now in the hook system. And a future-looking suggestion (consideration, really) to further unifiy sim backends vs. components/systems.

src/holosoma/holosoma/simulator/isaacgym/isaacgym.py
juan-g-bonilla: i considered having a `START` phase or something like that, but I struggled to find a practical difference between a hook on `START` vs just doing your logic on `__init__` of the hook instance (which get called in PR#160 at the end of simulator init and before the sim runs).

We could add artificial phase breaks to resolve some of these dependencies between plugins (gantry is created, then a plugin is setup to control gantry):
```
self.hook_manager.emit(Phase.SETUP)
self.hook_manager.emit(Phase.START) # can depend on behavior in SETUP
```

we have a similar problem with sensors + sensor publishing. Should we have?:
```
self.hook_manager.emit(Phase.RENDER_CAMERAS)
self.hook_manager.emit(Phase.TELEMETRY)
```

but that's baking in plugin behavior into the simulator time stepping. As far as the simulator understands, there's no difference between `RENDER_CAMERAS` and `TELEMETRY`, it's all really `POST_PHYSICS`. Similarly, there's no difference between `SETUP` and `START`, it's all "i'm done initializing and I'm about to play". It feels like we're poor-manning a dependency logic by arbitrarily splitting phases. 

That being said, other tools do usually have a `SETUP` and a `START`, at the very least for convenience.

src/holosoma/holosoma/simulator/base_simulator/hooks.py
juan-g-bonilla: that would be better. i only kept the for loop for historic reasons. Currently:
```
    def reset_envs_idx(self, env_ids, target_states=None, target_buf=None):
        """Reset some environments and handle video recording callbacks."""

        # Call episode end for environments that are being reset
        for env_id in env_ids:
            if hasattr(self.simulator, "on_episode_end"):
                self.simulator.on_episode_end(env_id.item())
```
so we'd need the simulator subclasses (which im not sure if there are any downstream user implementers) to adapt their `on_episode_end` to take in a tensor.

src/holosoma/holosoma/simulator/shared/video_recorder.py
```diff
@@ -98,6 +97,18 @@ def __init__(self, config: VideoConfig, simulator: BaseSimulator) -> None:
...
+    def register_hooks(self, hooks: HookRegistry) -> None:
+        """Register video lifecycle hooks with the simulator loop.
+
+        Frames are captured on FRAME_END, which fires once per frame (after the last physics substep +
+        tensor refresh) — the control rate the recorder wants — so no manual decimation against
+        control_decimation is needed.
+        """
+        hooks.add(Phase.EPISODE_START, self.on_episode_start, name="video.on_episode_start")
+        hooks.add(Phase.EPISODE_END, self.on_episode_end, name="video.on_episode_end")
+        hooks.add(Phase.FRAME_END, self.capture_frame, name="video.capture_frame")
```
samuelgundry, line 109: Bug?  capture_frame expects `env_id` but `FRAME_END` does not pass?
How can we catch these static/typing errors without eyeballs or runtime?

src/holosoma/holosoma/utils/sim_utils.py
```diff
@@ -488,8 +490,15 @@ def run(self) -> None:
...
-                # Direct simulator step - this triggers bridge.step() inside simulate_at_each_physics_step()
+                # Direct simulator step with the same physics hooks used by BaseTask.
+                if step_count % control_decimation == 0:
+                    self.simulator.hooks.emit(Phase.FRAME_BEGIN)
+                self.simulator.hooks.emit(Phase.PRE_STEP)
                 self.simulator.simulate_at_each_physics_step()
+                self.simulator.hooks.emit(Phase.POST_STEP)
+
+                if step_count % control_decimation == 0:
+                    self.simulator.hooks.emit(Phase.FRAME_END)
```
samuelgundry, line 501: I think very slight behavior change means VideoRecorder will now capture at step_count=0 (before step_count is incremented) whereas before it captured only after incrementing step_count (modulo decimation)?  See lines 198 removed in video_recorder.py.   So for `control_decimation=4`,  `FRAME_END` now also fires on on step 0 vs. before only on step 4, 8, ...  Not sure if this is intentionally?

juan-g-bonilla: it's not intentional, claude also brought this up initially when prompted about behavior change:
```
OLD run_sim: captures at steps [3, 7, 11, ...]
NEW run_sim: captures at steps [0, 4, 8, 12, ...]
```

I couldn't find a simple fix and the difference in behavior was subtle enough (and nothing really depends quantitatively on a video recording, it's just logging) that I didn't think it was worth to fix. Lmk if you feel otherwise.

src/holosoma/holosoma/simulator/shared/video_recorder.py
juan-g-bonilla: It's not a bug, although it does read a bit weird. I actually never touched the implementation of `capture_frame`, which was already:
```
    def capture_frame(self, env_id: int = 0) -> None:
        # Only capture frames when recording is active and from the correct environment
        if not self._is_recording or env_id != self.config.record_env_id:
            return
```

so `FRAME_END` doesn't pass `env_id`, but that's ok because it's defaulted on the method signature. However, if the passed value is not the `record_env_id`, it will be noop. Overall, I think is confusing behavior and `capture_frame` should not take any value and directly use `self.config.record_env_id`. However, in the spirit of keeping this PR completely non-API changing, I didnt do these kinds of cleanups.

On:
>  How can we catch these static/typing errors without eyeballs or runtime?

I added commit 2b0597e63a1b6538fb23f0462d2b36e8f40b51f1 which includes more early checking mechanisms. I added python `@overload` to the `add` function for the specific phases so that the IDE, mypy, and hopefully agents will complain about using the wrong callable for each phase. Also added an `inspect.signature` check on `add`.

samuelgundry: I see, yeah, right, `end_id` arg is ignored for `record_env_id`. Yes, agree leave existing as-is for now. thanks for clarifying.

samuelgundry (approved):

PR #159, opened 2026-07-10
unify config presets under a self-registering ConfigRegistry

juan-g-bonilla:
## Summary

Replaces the per-module `DEFAULTS` dicts + ad-hoc plugin/menu loading with one self-registering **`ConfigRegistry`** per config family, shared across the training (`holosoma`) and inference (`holosoma_inference`) packages. Entry-point group names and the `ep.load()`-returns-value contract are unchanged.

To the best of my knowledge **this is not a breaking change**: the `module.DEFAULTS` and `get_defaults()` shims still resolve to the live registry, so existing extensions keep working.

## Features

- **Self-declaring presets — one object, one place.** `.add()` names a preset *and* returns it, so the module attribute and the menu entry can't drift. Replaces the config-plus-hand-maintained-`DEFAULTS`-dict-plus-separate-`menus`/`plugins`-layer dance. `.add()` is type-checked, so a mistyped preset fails at import, not as a confusing CLI error later.

  ```python
  # BEFORE — value here, name registered separately (keep in sync by hand)
  disabled = DisabledLoggerConfig()
  DEFAULTS = {"disabled": disabled, "wandb": wandb}

  # AFTER — the registry IS the source of truth
  LOGGER_REGISTRY = ConfigRegistry((DisabledLoggerConfig, WandbLoggerConfig), group="holosoma.config.logger")
  disabled = LOGGER_REGISTRY.add("disabled", DisabledLoggerConfig())
  ```

- **`--import-file` — try a preset without packaging anything.** Point the CLI at a local Python file that registers presets. No `pip install`, no entry point, no repo edit — ideal for one-off experiments.

  ```python
  # my_presets.py (anywhere on disk)
  from holosoma.config_values.robot import ROBOT_REGISTRY, g1_29dof
  ROBOT_REGISTRY.add("g1_stiff", replace(g1_29dof, control=replace(g1_29dof.control, action_scale=0.1)))
  ```
  ```bash
  python -m holosoma.run_sim --import-file my_presets.py robot:g1-stiff   # selectable like a built-in
  ```

- **Dynamic `dict[str, X]` CLI — build a keyed mapping from the command line.** `<field>.<key>:<variant>` tokens populate a registry-backed dict, each value drawn from a menu and independently overridable. Unknown variants fail loud with the valid choices.

  ```bash
  python -m holosoma.run_sim sensors.front:rgb sensors.wrist:depth --sensors.front.width 1280
  #   sensors.front:potato -> "Unknown sensors variant 'potato'; choose from ['depth', 'rgb']"
  ```

- **`UseRegistry` — explicit field binding.** Fields say which menu they draw from in the annotation. Re-read at parse time, so entry-point and `--import-file` presets appear in the menu automatically.

  ```python
  simulator: Annotated[SimulatorConfig, UseRegistry(RUN_SIM_REGISTRY)] = ...
  robot:     Annotated[RobotConfig,     UseRegistry(ROBOT_REGISTRY)]   = ...
  ```

- **One code path for every tool.** `parse_config()` is the single entry: extract `--import-file` → `load_plugins()` (import built-ins + ingest entry points) → populate dynamic dicts → `tyro.cli`. `run_sim` / `train_agent` / `eval_agent` / `replay` / `run_policy` all route through it, so a preset registered any one way is visible from every tool.


src/holosoma/holosoma/utils/config_registry.py
```diff
@@ -0,0 +1,285 @@
+"""Registry helpers for named config presets."""
```
samuelgundry, line 1: Not blocking/not a biggie: it'd be nice to have the "framework" code in a config-specific module rather than alongside the user-facing config values.  Perhaps config_types/ but even then it's user-facing, so we don't really have a place today.

src/holosoma/holosoma/utils/pycompat.py
```diff
@@ -0,0 +1,22 @@
+"""Python version compatibility shims.
```
samuelgundry, line 1: "compat.py" as a name isn't great, perhpas pycompat.py  or something tying to Python.

src/holosoma_inference/holosoma_inference/utils/config_registry.py
```diff
@@ -0,0 +1,301 @@
+"""Registry helpers for named config presets."""
+
+# NOTE FOR AGENTS/MAINTAINERS: This module is intentionally mirrored with
+# holosoma.config_values.registry. Keep behavior changes in sync unless a difference is
+# package-specific.
```
samuelgundry, line 5: Yup... One day we'll figure an acceptable shared library for utils, etc.

samuelgundry (dismissed):
Good stuff. It's a big step towards improving config maintainability and making it easier to add/extend/refactor.   The backwards compatibility + deprecation path is a GOOD thing, thank you. We'll need to keep that there and potentially  consider a "version bump" to remove entirely later.

samuelgundry (approved):

PR #160, opened 2026-07-11
Plugin config family (PLUGIN_REGISTRY) + ROS2 clock/gantry/odometry plugins

juan-g-bonilla:
Builds on the lifecycle hook registry (base branch **#158**): the step loop emits phase events, and behavior attaches as callbacks. Core features (video recording, simulator bridge, virtual gantry) already run this way. This PR lets **config/CLI attach the same kind of behavior** without subclassing a backend, and ships the ROS2 example plugins.

## What this adds

- A `PluginConfig` family (`PLUGIN_REGISTRY`, group `holosoma.config.plugin`) selectable as the dynamic-dict CLI field `plugin.<key>:<preset>`.
- A **plugin** is any class constructed as `cls(cfg, simulator)` that registers hooks on `simulator.hooks` in `__init__` — **duck-typed, no base class**. It pairs with a `PluginConfig` whose `get_cls()` returns the runtime class. Same registry the core features use.
- Plugins are instantiated in `BaseSimulator.__init__` (from `FullSimConfig.plugin`), so their hooks are registered before the loop emits any phase.
- Threaded through **both** entry points: `RunSimConfig.plugin` (sim2sim) **and** `ExperimentConfig.plugin` → `base_task` → `FullSimConfig` (training).
- Shipped presets: `none` (no-op), `clock_publish` + `gantry_control` (ROS2), and `odometry` — a self-sourced (non-camera) egress plugin that reads `robot_root_states` each control step and publishes `nav_msgs/Odometry`.

Why: one mechanism for step-loop side effects, core and custom alike — added logging, ROS2 bridges, teleop, egress, etc. become config presets (packaged entry point or `--import-file`) instead of loop forks.

## API changes

- **New field `FullSimConfig.plugin: dict[str, PluginConfig]`** (default `{}`), plus registry-bound **`RunSimConfig.plugin`** and **`ExperimentConfig.plugin`** (and `EnvConfig.plugin`, threaded through `get_tyro_env_config`). Additive — existing configs unaffected.
- **New public `PluginConfig` ABC** (`holosoma.config_types.plugin`): subclass with your knobs as dataclass fields + implement `get_cls()`. Also `NoOpPluginConfig`, `ClockPublishPluginConfig`, `GantryControlPluginConfig`, `ROS2OdometryPluginConfig`.
- **New entry-point group `holosoma.config.plugin`** for packaged plugin presets.
- **New extra `holosoma[ros2]`** — opt-in marker for the ROS2 plugins (install `rclpy` from your ROS2 distro / RoboStack; the impls import it lazily so core stays importable without ROS).

## Behavior changes

- **None by default.** The `plugin` field is empty unless you select plugins, so no existing run changes.
- When selected, a plugin's hooks fire at lifecycle phases (from the base branch): `FRAME_BEGIN`, `PRE_STEP`, `POST_STEP`, `FRAME_END` (periodic), `EPISODE_START` / `EPISODE_END`, `CLOSE` (once, reverse order).
- Rate fields accept an int decimation **or** a frequency string (`"100Hz"`, `">100Hz"`, `"<100Hz"`), resolved against the phase's base rate; the registry sub-samples natively.

## Usage

```bash
# ROS2: publish /clock at 100 Hz + drive the gantry over three topics
python -m holosoma.run_sim plugin.clk:clock_publish --plugin.clk.publish-every=100Hz plugin.g:gantry_control

# ROS2 odometry: publish the robot base pose/velocity as nav_msgs/Odometry
python -m holosoma.run_sim plugin.odom:odometry --plugin.odom.topic=/odom
```

Write-your-own (config + duck-typed class), phases, and cadence are documented in `docs/writing-extensions.md` (Plugins section).

---
_Stacked on #158 (lifecycle hook system). Review after / alongside it; the diff here is the plugin family only._

### DIFF
--- a/src/holosoma/holosoma/simulator/base_simulator/base_simulator.py
+++ b/src/holosoma/holosoma/simulator/base_simulator/base_simulator.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from loguru import logger
 
@@ -175,6 +175,14 @@ class BaseSimulator:
         self.debug_viz_enabled = self.simulator_config.debug_viz
         self.object_registry = ObjectRegistry(device)
         self.hooks = HookRegistry(base_rates=self._hook_base_rates())
+        # Build plugins from tyro_config.plugin and keep the instances alive (key -> plugin).
+        # Constructing each here registers its hooks on self.hooks, so they fire on later
+        # emit(). The `none` preset disables a slot.
+        self.installed_plugins: dict[str, Any] = {
+            key: cfg.get_cls()(cfg, self) for key, cfg in tyro_config.plugin.items()
+        }
+        if self.installed_plugins:
+            logger.info(f"Installed plugins: {sorted(self.installed_plugins)}")
         self._closed = False
 
         # Virtual gantry system
--- a/src/holosoma/holosoma/simulator/shared/ros2_plugins.py
+++ b/src/holosoma/holosoma/simulator/shared/ros2_plugins.py
@@ -1,10 +1,12 @@
 """ROS2 example plugins.
 
-Two reference plugin implementations that talk ROS2 (each constructed as
+Reference plugin implementations that talk ROS2 (each constructed as
 ``cls(cfg, simulator)`` and registering hooks on ``simulator.hooks`` — no base class):
 
 - :class:`ClockPublishPlugin` publishes sim time as ``rosgraph_msgs/msg/Clock``.
 - :class:`GantryControlPlugin` drives the virtual gantry from three independent topics.
+- :class:`ROS2OdometryPlugin` publishes the robot base pose/velocity as ``nav_msgs/Odometry`` — a
+  self-sourced egress that reads ``robot_root_states`` each control step (no camera frames).
 
 rclpy and the ROS message packages are an **optional** dependency (``holosoma[ros2]``).
 They are imported lazily inside methods (never at module top), mirroring
@@ -26,7 +28,11 @@ from loguru import logger
 from holosoma.simulator.base_simulator.hooks import Phase
 
 if TYPE_CHECKING:
-    from holosoma.config_types.plugin import ClockPublishPluginConfig, GantryControlPluginConfig
+    from holosoma.config_types.plugin import (
+        ClockPublishPluginConfig,
+        GantryControlPluginConfig,
+        ROS2OdometryPluginConfig,
+    )
     from holosoma.simulator.base_simulator.base_simulator import BaseSimulator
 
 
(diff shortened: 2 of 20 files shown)


src/holosoma/holosoma/simulator/shared/ros2_plugins.py
```diff
@@ -0,0 +1,206 @@
...
+        if node is not None:
+            node.destroy_node()
+            self._node = None
+
+
+class GantryControlHook:
+    """Control the virtual gantry over ROS2 via three independent topics.
+
+    Each property is its own subscription, so publishing to one topic changes only that
+    property:
```
samuelgundry, line 103: Let's sync offline for this design/architecture pattern, as it's a concrete example of timing/ordering of dependencies related to my comment in #158 regarding init()/register() ordering. GantryControlHook depends on VirtualGantry, but installed plugins are initialized before the bespoke gantry, video recorder etc. Curious what moving gantry and video recorder into this plugin system would look like -- at least as a way to test the design here and dependencies between plugins/hooks.

src/holosoma/holosoma/config_types/hook.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@dataclass(frozen=True)
+class HookConfig(abc.ABC):
+    """Base config for a simulator hook plugin.
+
+    Subclass with the plugin's parameters as dataclass fields and implement
+    :meth:`get_cls` to point at the runtime hook class. Register an instance in
+    ``HOOK_REGISTRY`` so it is selectable as ``hook.<key>:<variant>`` on the CLI.
+    """
```
samuelgundry: I actually think "Plugin" is the more apt and intuitive name. :)    And Plugins depend on the hook system, as well as other contracts/mechanisms.  A Plugin is a set of hooks + behaviors.

src/holosoma/holosoma/config_types/hook.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def get_cls(self) -> Callable[..., Any]:
+        from holosoma.simulator.shared.builtin_hooks import NoOpHook
+
+        return NoOpHook
+
+
+@dataclass(frozen=True)
+class LogRobotStateHookConfig(HookConfig):
+    """Reference hook config: periodically log the robot base pose.
```
samuelgundry: I really dislike single config_types/hook.py for everything. It's mixes framework (`HookConfig`) with and forces coupling of otherwise unrelated user-facing code but anyways, it is what it is with our current config system.

src/holosoma/holosoma/simulator/base_simulator/base_simulator.py
```diff
@@ -175,6 +175,12 @@ def __init__(self, tyro_config: FullSimConfig, terrain_manager: TerrainManager,
         self.debug_viz_enabled = self.simulator_config.debug_viz
         self.object_registry = ObjectRegistry(device)
         self.hooks = HookRegistry(base_rates=self._hook_base_rates())
+        # Build hook plugins from tyro_config.hook and keep the instances alive (key -> hook).
+        # Constructing each here registers its callbacks on self.hooks, so they fire on later
+        # emit(). The `none` preset disables a slot.
+        self.installed_hooks: dict[str, Any] = {key: cfg.get_cls()(cfg, self) for key, cfg in tyro_config.hook.items()}
+        if self.installed_hooks:
+            logger.info(f"Installed hook plugins: {sorted(self.installed_hooks)}")
         self._closed = False
```
samuelgundry: Will sync offline: my main (relatively minor) concern is ordering of init hooks/plugins and when/how they might need to know about each other, or other parts of the simulator backend. GantryControl needs to check for simulator.virtual_gantry existence, whereas an extra phase or two in the hook system eliminates this incidental coupling by encoding it into hook callbacks (e.g OnReady/OnBegin or whatever are intuitive names for this domain).

samuelgundry (commented):
It's great to have examples that demonstrate the power of the hook-plugin system. Per comments below, I'll sync first with you and brain storm whether we want refinements now or deferred.

src/holosoma/holosoma/config_types/hook.py
juan-g-bonilla: Plugins it is. Believe it or not, I had a 20 min chat with claude about the semantics for this. `extension`, `bridge`, `hook`, `telemetry`, `behavior`... I think `plugin` we rejected because, to me, it implied "behavior outside of core holosoma", while we want this system to also be used extensively within this repo. That being said, using plugin both for "in-house" and "out-house" systems reinforces how we're not privileged users of our tool and any behavior we can implement in-house should be possible out of it. I'm a fan of mitsuba and they follow [this pattern](https://mitsuba.readthedocs.io/en/stable/src/plugin_reference.html), so I'm happy to just do plugin.

samuelgundry (approved):

PR #168, opened 2026-07-15
fix: remove invalid config= kwarg in inference service_node

asetapen:
The inference service node's `main()` called parse_config with a `config=` keyword argument that collided with the function's own first positional parameter (also named `config`), making the entrypoint crash on startup. This removes the bad `kwarg` and its now-unused import.

This surfaced after #159 reworked the config-parsing path; the `run_policy.py` call sites were updated to the new signature but this service-node call site was left with the old-style kwarg.


juan-g-bonilla:
Thanks for catching this!

juan-g-bonilla (approved):

juan-g-bonilla:
fixed in:  https://github.com/amazon-far/holosoma/pull/167

PR #171, opened 2026-07-16
[retargeting] per-window z_floor, per-frame object points, headless data conversion

Juyue:
feat: per-window z_floor, per-frame object points, headless data conversion

- Add per-window z_floor override to FootLockConfig (3-tuple variant)
- Support per-frame object points in interaction mesh retargeter
- Add headless mode to data conversion for batch processing


lijiaman (approved):
Looks good.

PR #170, opened 2026-07-15
Add camera sensor simulation 

juan-g-bonilla:
# PR — `feat(sensors): backend-agnostic mounted cameras + camera-egress plugins`

## Summary

Adds a **backend-agnostic mounted-camera system** and the **camera-egress plugins** that ship rendered
frames off the sim. Declare any number of cameras in one top-level `sensor:` config; mount them on a
robot link, a scene actor, or the world; render RGB and/or depth on all four backends (MuJoCo Classic
CPU, MuJoCo Warp GPU, IsaacGym, IsaacSim); read frames through one unified `get_camera_data` API with a
shared cross-device cache; feed them to visual-policy **observation terms**; and stream them out via
**plugins** (`plugin.<key>:ros2-image` / `viz` / …) — ROS2 `Image`/`CompressedImage` publish, a live cv2
window, or an mp4. Built on the plugin family (base branch): camera egress is a plugin like any other, so
it works in both `run_sim` and training. This is the sensor/vision layer of the stack.

### Newly supported (opt-in; no-op default — no existing task/env affected)

- **`sensor:` camera config family, selectable as `--sensor.<name>:<variant>`.** New
  `CameraSensorConfig` (`config_types/sensor.py:...`) — backend-agnostic core fields
  (width/height/vertical_fov/near/far/data_types/update_decimation) + a `SensorMountConfig`
  (`robot_link` / `actor` / `world`) + per-backend sub-configs — resolved from `CAMERA_REGISTRY`
  (`config_values/sensor.py:27`, group `holosoma.config.sensor`). Surfaced as a dynamic-dict field:
  `sensor` on `RunSimConfig` (`run_sim.py:83`) / `ExperimentConfig` (`experiment.py:129`) /
  `EnvConfig` (`env.py:30`), threaded into `FullSimConfig.sensors` (`full_sim.py:34`, plural). Empty by
  default. G1 building-block cameras (`g1-head`, `g1-stereo-left/right`, `g1-*-wrist`, `g1-waist-*`)
  ship in `config_values/wbt/g1/sensor.py` — composed per-key on the CLI.
- **Per-backend camera rendering, unified read.** Each backend renders due cameras into a
  `SensorManager`/`CameraRuntime` registry (`simulator/shared/camera_sensor.py`) with a lazy
  cross-device cache; frames are read back through `BaseSimulator.get_camera_data(name, modality,
[...]

PR #172, opened 2026-07-16
fix(isaacsim): disable DLSS AA so camera-sensor tests don't hang the docker-build gate

juan-g-bonilla:
## Problem

The `main` **Docker Build** run for #170 failed its blocking `Gate: isaacsim tests (sha image)` ([run 29478427690](https://github.com/amazon-far/holosoma/actions/runs/29478427690/job/87568597321)). All **11 isaacsim camera-sensor tests** (new in #170) hit `subprocess.TimeoutExpired` at 700s each and were SIGKILL'd (`returncode: -9`), ballooning the gate from ~12m (its historical runtime) to **2h41m** and blocking `:latest` promotion for every image.

The subprocesses were **not** crashing or failing an assertion — they boot IsaacSim fully and hang. The last line of captured output before every hang is:

```
[rtx.postprocessing.plugin] DLSS increasing input dimensions: Render resolution of (37, 37) is below minimal input resolution of 300.
```

(These tests never passed before: #170's own `pytest.yaml` isaacsim lane is `continue-on-error: true`, so the same 2h35m hang was masked as "success". This blocking docker-build gate is simply the first place they had to pass.)

## Root cause

IsaacLab's `RenderCfg` defaults `antialiasing_mode` to **DLSS**. DLSS is an AI upscaler intended for large viewports; on the tiny (64×64) perception cameras these tests use it is meaningless (hence the "below minimal input resolution of 300" warning), and on the **A10G CI runner** its init blocks indefinitely in a native call — hanging every `TiledCamera` render. The isaacsim **training** smoke tests in the same job passed because they run with `enable_cameras=False`: no RTX render products are created, so DLSS never engages.

## Fix

Set `SimulationCfg.render = RenderCfg(antialiasing_mode="Off")` (one construction site; covers the whole backend). No AA is correct for these cameras regardless — there is nothing to upscale — and it keeps camera rendering deterministic across GPUs.

## Verification

Reproduced and fixed on an L40S dev box (`hssim` env, IsaacSim 5.1 / IsaacLab 2.3):

- **Before:** default DLSS — camera geometry test passes on the L40S but emits the DLSS warning (the A10G is where it hangs).
- **After:** `antialiasing_mode="Off"` — `camera_geometry_assert.py --simulator isaacsim` still prints `env0: geometry OK` / `PASS`, writes `OK` to its result-file, **and the DLSS warning is gone** from the log.

The full validation across all 11 camera scenarios runs on the A10G in this PR's isaacsim gate.

### DIFF
--- a/src/holosoma/holosoma/simulator/isaacsim/isaacsim.py
+++ b/src/holosoma/holosoma/simulator/isaacsim/isaacsim.py
@@ -33,7 +33,7 @@ from isaaclab.sensors import (
     TiledCameraCfg,
     patterns,
 )
-from isaaclab.sim import PhysxCfg, SimulationCfg, SimulationContext
+from isaaclab.sim import PhysxCfg, RenderCfg, SimulationCfg, SimulationContext
 from isaaclab.sim.utils import bind_physics_material
 from isaaclab.terrains import TerrainGeneratorCfg, TerrainImporterCfg
 from isaaclab.terrains.utils import create_prim_from_mesh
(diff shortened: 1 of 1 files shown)


juan-g-bonilla:
This was not what was triggering the CI to fail on rendering tests. Will open a new PR with the proper fix.

PR #173, opened 2026-07-16
bridge: publish base odometry from sim over the SDK (rt/odommodestate)

juan-g-bonilla:
Publish base odometry from the simulator over the SDK, so the sim emits the same base-state channel the real robot's onboard sport/loco mode does (`SportModeState` on `rt/odommodestate`). A downstream telemetry `read_odom_state -> /telemetry/odom` is then identical in sim and on hardware.

## What
- `BasicSdk2Bridge._get_base_odometry` — simulator-agnostic read of the unified `robot_root_states` `[pos, quat_xyzw, lin_vel_world, ang_vel_world]`, rotating world→body velocity via `quat_rotate_inverse` and converting quat xyzw→wxyz. Default `publish_odom()` is a no-op, so SDKs without a base-state channel (booster) do nothing.
- `UnitreeSdk2Bridge.publish_odom` fills an `OdomState` and calls `interface.publish_odom_state`.
- `UnitreeMpSdk2Bridge.publish_odom` computes fields in the binding-free parent and ships plain floats to the DDS child via a new `publish_odom_state` RPC (mirrors `publish_low_state`, keeps CycloneDDS out of the rclpy process).
- `SimulatorBridge.step` calls `publish_odom` only when `BridgeConfig.publish_odom` is set (default `False`).

## Tests
`test_unitree_mp_bridge.py` round-trips odom through the spawned child: position pass-through, world→body velocity, xyzw→wxyz quat, yaw_speed. Runs under `no_sim`.

### DIFF
--- a/src/holosoma/holosoma/bridge/base/basic_sdk2py_bridge.py
+++ b/src/holosoma/holosoma/bridge/base/basic_sdk2py_bridge.py
@@ -314,6 +314,47 @@ class BasicSdk2Bridge(ABC):
 
         return quaternion, gyro, acceleration
 
+    def _get_base_odometry(self):
+        """Get base odometry: position, orientation, body-frame linear velocity, yaw rate.
+
+        Simulator-agnostic — reads the unified ``robot_root_states`` 13-vector
+        ``[pos(3), quat_xyzw(4), lin_vel_world(3), ang_vel_world(3)]``, the sim analog of the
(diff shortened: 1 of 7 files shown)


gabriellemerritt (approved):
Looks good to me, in another PR we should point to the correct wheel version that enables this

PR #182, opened 2026-07-30
harden(docker): apt upgrade in all Dockerfiles; inference image builds via uv

clayrosenthal:
Add apt-get upgrade -y before apt install in every Dockerfile so base-image packages pick up security fixes on rebuild. Also convert the inference image to build via uv (uv_build backend declared in pyproject.toml) instead of pip3.


clayrosenthal:
test [docker build run](https://github.com/amazon-far/holosoma/actions/runs/30584166304)

lindashih (approved):

PR #184, opened 2026-08-10
fix(augmentation): name-based gait-phase channel mirror for symmetry aug

tkevinbest:
## What

Symmetry-based data augmentation mirrors observations across the robot's sagittal plane. The gait-phase signals `sin_phase` and `cos_phase` are separate observation terms, each carrying one channel per foot, ordered to match `env.feet_indices`. Under a left-right mirror, each foot's phase channel must be **swapped** with its mirror foot's channel.

The old code did neither a full mirror nor a swap:

```python
sin_phase[..., 0] = -sin_phase[..., 0]
```

It only ever touched **channel 0** and **negated** it. So:

- Every foot channel other than channel 0 was left untouched — for a biped, the second foot's phase was never mirrored at all. This affects the standard left-first g1 case, not just unusual orderings.
- Negating channel 0 is not a consistent left-right leg swap even for that one channel.

The result is an inconsistent mirrored gait phase, which corrupts the symmetry-augmented transitions used in training (observed as a wide-stance gait when augmentation was enabled).

## Fix

Mirror the gait phase by **swapping** each foot's channel with its left-right mirror foot's channel, derived by name:

- For each foot channel, find the foot whose name is its left/right mirror and map the channel to that foot's index (in `env.feet_indices` order). This gives a permutation `p` where `mirrored_phase[..., i] = phase[..., p[i]]`. For g1 this resolves to `[1, 0]` — a genuine left/right swap.
- Mirror resolution uses a new helper `_mirror_body_name`, which swaps side tokens (`left`/`right`, and `l_`/`r_` prefixes or `_l_`/`_r_` infixes) only at word boundaries — so a mid-word letter like the `l` in `ankle_roll` is never misread as a side marker.
- If a channel's mirror can't be resolved by name, it falls back to identity (maps to itself), so behavior degrades to "unchanged" rather than "wrong".

Correct for any foot ordering and any number of feet, bipeds and quadrupeds.

## Behavior change

**This changes the mirrored phase for every robot, including the existing left-first biped case** — it is not a no-op refactor. Old vs new for a left-first biped:

- Old: `[-sin(φ0), sin(φ1)]` (negate channel 0, ignore channel 1)
- New: `[ sin(φ1), sin(φ0)]` (swap the two feet)

The new behavior is the intended consistent leg swap. Any policy trained with symmetry augmentation on the old code was augmenting against a corrupted phase mirror.

## Scope

Single file: `agents/modules/augmentation_utils.py` (+63 / −23). Isolated to the augmentation path.

## Validation (g1 PPO)

Matched A/B on `exp:g1_29dof`: same machine, seed 42, 4096 envs, 5000 iters; control at `bccd4d7`, fix at `8f17e55`.

**Equivariance is fixed.** On-policy residual `pi(M s) - M_a pi(s)`, both policies scored against the corrected mirror, over 384k visited states:

| | control | fix |
|---|---|---|
| overall | 0.0278 | 0.0054 |
| standing | 0.0502 | 0.0032 |
| moving | 0.0211 | 0.0058 |
| ankle pitch (L / R) | 0.0423 / 0.0429 | 0.0044 / 0.0044 |

Every joint improves. Ankle pitch is ~10x lower and left/right balanced. Scoring both policies on a shared state buffer gives the same result (5.1x on control states, 5.4x on fix states), so this is not a state-distribution effect.

**There is a task performance gap.** Same A/B, final values:

| | control | fix |
|---|---|---|
| `rew_tracking_lin_vel` | 0.880 | 0.694 |
| `rew_tracking_ang_vel` | 0.993 | 0.905 |
| `rew_feet_phase` | 4.301 | 4.038 |
| mean episode length | 999 | 919 |

### DIFF
--- a/src/holosoma/holosoma/agents/modules/augmentation_utils.py
+++ b/src/holosoma/holosoma/agents/modules/augmentation_utils.py
@@ -188,6 +188,56 @@ class SymmetryUtils:
             dtype=torch.float,
         )
 
+        # Build a NAME-BASED foot/leg channel mirror map for gait-phase observations.
+        # Gait phase obs (sin_phase/cos_phase) have one channel per foot, ordered to
+        # match env.feet_indices == [body for body in body_names if foot_body_name in body].
+        # Under a left-right mirror, each foot's phase channel must be SWAPPED with its
+        # mirror foot's channel. The previous implementation hard-coded "negate channel 0"
+        # which only happens to be (partially) correct when channel 0 is the LEFT foot
+        # (i.e. left-first body_names). This derives the permutation by name so it is
+        # correct for any foot ordering and any number of feet (bipeds and quadrupeds).
+        self.phase_channel_map = self._build_phase_channel_map()
+
+    def _mirror_body_name(self, name: str) -> str:
+        """Return the left<->right mirror of a body name. Side tokens are matched only
+        at WORD BOUNDARIES (name start, or surrounded by '_') so mid-word letters are
+        never mistaken for a side prefix -- e.g. the 'l' in 'ankle_roll' must NOT be
+        read as an 'l_' side marker. Returns the name unchanged if no side token found.
+        """
+        import re
+
+        # whole-word 'left'/'right'
+        if re.search(r"(?<![a-z])left(?![a-z])", name) and "right" not in name:
+            return re.sub(r"(?<![a-z])left(?![a-z])", "right", name, count=1)
+        if re.search(r"(?<![a-z])right(?![a-z])", name) and "left" not in name:
+            return re.sub(r"(?<![a-z])right(?![a-z])", "left", name, count=1)
+        # prefix l_/r_ (start) or _l_/_r_ (after an underscore) ONLY
+        if re.match(r"^l_", name):
+            return "r_" + name[2:]
+        if re.match(r"^r_", name):
+            return "l_" + name[2:]
+        if "_l_" in name:
+            return name.replace("_l_", "_r_", 1)
+        if "_r_" in name:
+            return name.replace("_r_", "_l_", 1)
+        return name
+
+    def _build_phase_channel_map(self) -> torch.Tensor:
+        """Permutation tensor p such that mirrored_phase[..., i] = phase[..., p[i]].
+
+        p[i] = index of the foot channel that is the left-right mirror of channel i.
+        Falls back to identity for any channel whose mirror can't be resolved by name
+        (so behavior degrades to "unchanged" rather than wrong)."""
+        body_names = list(getattr(self.env, "body_names", []) or [])
+        foot_pattern = self.robot_config.foot_body_name
+        foot_names = [b for b in body_names if foot_pattern in b] if foot_pattern else []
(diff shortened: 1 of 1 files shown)


tkevinbest:
**Test, and why it doesn't rely on the trainer's mirror.** Feed the policy the canonical mirror-symmetric standing observation: zero `dof_pos`, `dof_vel`, previous action; projected gravity `[0, 0, -1]`; zero velocity command; both gait-phase channels at pi (what a stand command produces). That vector equals its own mirror, so a symmetric policy must satisfy `a[i] == sign[i] * a[perm[i]]`. Any residual is unforced policy asymmetry. `Loss/symmetry_actor_loss` is `MSE(pi(T s), T_a pi(s))` with `T` the trainer's own mirror, so a wrong `T` makes it self-referential: it converged to 2.9e-5 on a policy this test shows is asymmetric. Ankle pitch reads most directly — the mirror leaves its sign unchanged, so symmetric means left equals right.

**29-DoF biped, PPO, `symmetry_actor_coef` 1.0**

| measurement | before | after |
|---|---|---|
| ankle-pitch L−R, canonical state | 0.114 | 0.0003 |
| equivariance residual RMS, 29 joints, logged rollout obs | 0.023–0.027 | 0.0035–0.0073 |
| same, ankle pitch only | 0.05–0.07 | 0.001–0.005 |

0.114 action units is ~1.6 deg of commanded target, ~1.3 Nm at that joint's stiffness.

**G1, today's nightly checkpoints (trained on main), same canonical test**

| policy | ankle-pitch L−R | RMS, 29 joints | largest residual |
|---|---|---|---|
| PPO, `symmetry_actor_coef` 1.0 | −0.0422 | 0.0398 | waist roll 0.144 |
| fast_sac, augmentation only | −0.0857 | 0.0460 | hip pitch ±0.116 |

The 29-DoF biped before the fix was RMS 0.0376, so the bias is not robot-specific.

**Measured hardware consequence.** Standing trials, 5 per policy, varied initial base poses. The fixed policy commands ~17 deg of left/right ankle-pitch target difference vs ~6 deg unfixed. With both soles flat, ankle pitch is kinematically pinned, so the command appears as opposing PD tracking error: ~14 Nm of opposing ankle-pitch torque, ~3.8x the summed ankle-pitch torque magnitude. Static equilibrium closes (net moment about the CoM under 3 Nm, vertical-axis moment under 0.8 Nm), so this is centre-of-pressure redistribution, not internal bracing. Mechanism: a constant action must lie in the +1 eigenspace of the action mirror, so an equivariant policy cannot hold a constant antisymmetric offset. The old code permitted one, and on a machine that is not itself mirror-symmetric that offset was partly cancelling a real asymmetry.

**Code note.** `_build_phase_channel_map` derives the permutation from `robot_config.foot_body_name`, but the phase channel is bound to physical legs through `foot_height_name` (via `feet_heights` in the `feet_phase` reward). Both filter the same `body_names` list and agree for the robots checked, so the permutation is correct here. A robot whose two filters order differently would get a wrong permutation, and the identity fallback would mask it.

PR #189, opened 2026-08-13
Harden general whole-body tracker training

noahcao:
## Summary

  - add simulator-native implicit position actuation for IsaacSim while preserving explicit torque control by default
  - add configurable preprocessing hooks and W&B artifact helpers
  - harden multi-motion loading, validation, sharding, transitions, velocity recomputation, and adaptive sampling
  - terminate and hold at source clip boundaries to prevent cross-clip teleports
  - synchronize PPO observation normalization once per rollout using merged Welford moments

  ## Testing

  - pre-commit hooks: passed
  - focused tracker, actuator, preprocessing, and normalization tests: 40 passed


carlosferrazza:
Quick question: `motion_end` termination is now the default for all g1 WBT training. Is this intended? Previously, we adopted original BeyondMimic reset, which restarts the episode and keeps going until episode length

PR #199, opened 2026-09-04
fix(inference): print WBT controls for policy subclasses

tkevinbest:
`_print_control_guide` identified WBT runs by exact class name:

```python
is_wbt = policy_class.__name__ == "WholeBodyTrackingPolicy"
```

WBT variants are registered under the `holosoma.policies.wbt` entry-point group, and those are subclasses. Running one printed the **locomotion** controls (`=`, `w/s`, `a/d`, `z`) and omitted `m - Start motion clip`, the key the operator needs to start the motion. `issubclass` fixes it.

Guide text only — no behaviour change. Two tests cover the subclass and locomotion branches; the first fails on `main`.


tomasz-lewicki (approved):
Makes sense, thanks for the PR!

PR #202, opened 2026-09-11
fix(wbt): correct velocities, body poses and clip boundaries in default-pose transitions

tkevinbest:
`MotionCommand` can splice a synthetic lead-in and lead-out between the robot's default pose and a retargeted clip, so a whole-body-tracking policy has something trackable at the clip boundary instead of a step change. It is gated by `enable_default_pose_prepend` / `enable_default_pose_append`, both default off, and no shipped config turns either on. That is also why the path had zero test coverage and seven defects.

What was wrong:

- Stored velocities were not the derivative of the stored positions. Position and velocity were each interpolated linearly between the endpoints, but the derivative of a linear position ramp is a constant, not a ramp. On the test fixture the stored value disagreed with the true derivative by up to 5.3 rad/s, sometimes with the wrong sign (+5.07 rad/s stored where the derivative was a constant -0.23 rad/s).
- Body poses in a frame were not forward-kinematics-consistent with that frame's joint angles. Blending each body's world position independently cuts an arc to a chord: link lengths shrank mid-transition, and no joint configuration corresponded to the emitted poses.
- With `motion_dir`, the transition was registered as its own motion interval. `step()` resamples a new clip the instant `time_steps` reaches a motion's end index, so the robot played the lead-in and was teleported to a random clip exactly at the handoff. Only the first and last clip of the concatenation got a transition at all.
- Motion body slots with no corresponding robot body were zero-filled, leaving a zero quaternion in those slots.
- A fake foot contact point aliases onto its ankle link, and both names are in the robot's body list, so the scatter into motion order was last-write-wins: the contact point, 3.7 cm below the ankle, overwrote the ankle. Both ankle links are tracked bodies, so this landed in the tracking reward.
- Body linear velocities were differenced from link-frame origins, but a motion file stores centre-of-mass velocities (the retargeting converter writes `body_pos_w` at the link origin and `body_lin_vel_w` at the CoM). Up to 16 cm of offset on this robot.
- Root rotation was slerped, which matches the clip's orientation at the seam but not its turn rate: a 0.97 rad/s angular-velocity step into a clip turning at 0.60 rad/s.

Joint angles and root position now come from a cubic Hermite segment that returns position and its analytic derivative from one polynomial, so the stored velocity is the derivative of the stored position by construction. Root rotation runs the same cubic in rotation-vector space, with the endpoint derivative mapped through the inverse left Jacobian of SO(3). Body poses are per-frame forward kinematics of the interpolated joint angles, batched across environments and clips rather than looped frame by frame. Body velocities are finite-differenced from those poses and referenced to each body's CoM. Every clip gets its own lead-in and lead-out, spliced into that clip's own frame interval, so `time_steps` walks from the transition straight into the clip's frames without a reset. The math lives in a new pure-math module, `utils/transition_trajectory.py`, with no env or simulator imports. FK stays IsaacSim-only, as before, and there are no new config knobs.

Both endpoints are at rest, which makes the cubic a smoothstep and the joint path monotone between the two poses. That is a deliberate trade. Matching the clip's actual arrival velocity would make the seam velocity-continuous, but it forces the path out past the target and back whenever the net displacement is small.

Verification: about 1500 of the 2400 added lines are tests. `utils/tests/test_transition_trajectory.py` covers the math (36 tests, no sim); `test_wbt_transition_fields.py`, `test_wbt_transition_splice.py` and `test_wbt_motion_scatter.py` cover the emitted fields, the splice and the body scatter under `no_sim`; and an IsaacSim harness asserts the root pose round-trips through FK, that distinct joint configurations give distinct body poses, and that the batched result matches looping one frame at a time. Batched-vs-per-frame agreement is 2.9e-06, which is what rules out the stale-cache failure mode, and link lengths hold to 2.4e-07 m while the root travels 64 cm. Full runs: `no_sim` 603 passed, IsaacSim 63 passed, pre-commit and mypy clean. On a 19 s 29-DoF clip (959 frames at 50 fps) the emitted joint angles arrive monotonically at the clip's first frame with no detour, against a stored velocity that is the derivative of the plotted position. A 300-iteration training run on this path completed with no NaNs.

Known limitations: feet can slide during the transition, and the default-pose root height still comes from the nominal spawn height rather than from FK, which is entangled with foot anchoring.


tkevinbest:
Ran the transition on a real clip (`motion_crawl_slope.npz`, 29-DoF, IsaacSim, A10G) with both directions enabled. The setup-time audit now reports:

```
prepend, over its 100 frames:
  max |joint_vel - d(joint_pos)/dt|      0.00073 rad/s
  max |body_lin_vel - d(body_com_pos)/dt| 0.00000 m/s
  seam step  0.1444 rad / 0.0196 m / 4.8035 rad/s
  clip's own frame-to-frame step reaches  0.1586 rad / 6.6987 rad/s

append, over its 100 frames:
  max |joint_vel - d(joint_pos)/dt|      0.00038 rad/s
  max |body_lin_vel - d(body_com_pos)/dt| 0.00000 m/s
  seam step  0.0757 rad / 0.0097 m / 3.3704 rad/s
```

The derivative contract holds on real data: body velocity is the exact difference of the emitted centre-of-mass positions, and the joint residual is the O(dt^2) truncation of the measuring difference, not an error in the stored values.

The seam numbers are worth reading against the scale on the last line -- the hand-off step is *smaller* than the clip's own motion between neighbouring frames (0.1444 vs 0.1586 rad, 4.80 vs 6.70 rad/s). A transition into a fast clip should step about as much as the clip does; this one steps slightly less.

Getting these numbers took one extra commit. The audit originally measured over the whole clip interval and printed a 7.6 m/s body residual on a healthy transition. Splitting the residual by region located the cause outside this change: 0.0000 m/s over the segment's own rows, with the 7.6 coming entirely from the clip's frames. A clip's `body_lin_vel_w` comes from `mj_objectVelocity` -- an analytic velocity from qvel -- while `body_pos_w` is a sampled position, so on fast motion the two disagree by metres per second with nothing wrong. The window now covers only the frames this code emits.

Still outstanding, unchanged from the description: the visual marker check. `replay.py` and `train_agent.py` both fail at CLI construction on the machine used, identically on `main` with the feature off (tyro 1.0.16 vs `config_types/randomization.py:68`), so the numbers above come from driving the env directly in Python rather than through the CLI.

tkevinbest:
Correcting my comment above: it is incomplete, and this PR has a defect. Do not merge yet.

That comment reported the segment as self-consistent, which it is -- velocity is the exact derivative of position, body poses are forward kinematics of the joint angles. But self-consistency is not the same as physical plausibility, and I had not checked the second thing. Checking it on a real clip:

```
endpoint A (default pose)      joint-limit violation  0.0000 rad
endpoint B (clip frame 0)      joint-limit violation  0.0260 rad
whole emitted segment          joint-limit violation  1.2411 rad
overshoot beyond the endpoints 2.1447 rad   <-- created by the interpolation
```

Worst joint: the two endpoints are 0.600 and 0.652 rad, a 0.05 rad move, but the segment swings to -1.545 rad through a -0.969 rad limit.

Cause: a cubic Hermite that must *arrive* at the clip's terminal velocity from rest will wind out and come back when the net displacement is small. Both bundled clips reach ~7.4 rad/s at frame 0, and the excursion scales with duration:

| duration | overshoot beyond endpoints |
|---|---|
| 0.25 s | 0.09 rad |
| 0.5 s | 0.23 rad |
| 1.0 s | 0.64 rad |
| 2.0 s (the default) | 1.71 rad |
| 4.0 s | 3.89 rad |

So a longer, gentler transition is worse, which is backwards, and the default sits in the bad region.

This is a regression on one axis relative to the code being replaced: the old linear interpolation was wrong about velocities, but its positions could never leave the endpoint envelope. Matching the terminal velocity is what buys the continuity this PR is about, and it is also what causes the excursion.

Marking this draft while I fix it. The fix will come with the check that was missing -- an assertion that no emitted frame exceeds the robot's joint limits -- so this class of defect cannot ship silently again. That assertion is the real gap: every existing test verified internal consistency, and none asked whether the result was reachable.

tkevinbest:
Fixed by capping the duration. On the clip that broke:

| | before | after |
|---|---|---|
| prepend segment limit violation | 1.2411 rad | **0.0260 rad** |
| append segment | 0.8202 rad | **0.0260 rad** |
| clip itself (untouched) | 0.1353 rad | 0.1353 rad |

0.0260 is exactly the clip's own frame-0 violation, so the segment inherits it and adds nothing of its own. The duration capped from 2.0 s to 0.4 s on the lead-in and 0.56 s on the lead-out, logged in both cases. Peak joint velocities still match the clip's 7.4 and 3.9 rad/s, so the seam continuity this PR is about is intact.

Why shortening rather than the alternatives: the conflict is not between matching position and matching velocity, it is created by duration. Arriving at speed from rest with little net displacement forces a wind-out, and the excursion grows with the duration -- 0.09 rad at 0.25 s, 1.7 rad at 2 s, 3.9 rad at 4 s. Shortening gives up neither property. Refusing to match the arrival velocity would leave a 7.4 rad/s step at the seam, against a clip whose own frame-to-frame steps reach 4.0 and sit at 0.9 for the 99th percentile -- a discontinuity well outside anything the clip contains. Scaling the velocity down is a smaller version of that same step.

So the configured duration is a ceiling now, not a target, and a silent correctness bug became a logged parameter adjustment. No knob added.

The more useful half is the check that was missing. Every test here verified that a frame's fields agree with each other; none asked whether the pose could be held. That is why a 2 rad excursion survived two review passes and a green suite. Emitted frames are now refused if they command a joint past its limit, and the guard is not stricter than the data it was handed -- a clip frame already outside its own limit stays the clip's problem.

Re-verified: `no_sim` 603 passed, FK gate passed, all pre-commit hooks pass. Audit on the real clip now reads 0.00000 m/s body residual and 0.019 rad/s joint residual, with seam steps below the clip's own frame-to-frame motion.

tkevinbest:
Draft again. The duration cap is not sufficient, found by testing a clip with a genuinely large pose gap.

Everything tested until now started close to the default pose -- the bundled clips and the one I was given are all within 0.7 rad and a few degrees of upright, so the transition had almost nothing to bridge. Two clips with real gaps behave very differently.

A large pose gap with a slow arrival works well:

```
pose gap 2.42 rad, root 61 cm below default, 66 deg tilt, arrival 0.18 rad/s
  prepend kept the full 2.0s, append capped to 1.28s
  joint residual 0.00025 rad/s, body residual 0.00000 m/s
  seam step 0.0036 rad / 0.19 rad/s, against the clip's own 0.93 rad / 26.2 rad/s
  link length sd 2.4e-07 m while the root travels 0.139 -> 0.781 m
```

A large pose gap with a fast arrival does not:

```
pose gap 2.49 rad, root 43 cm below default, 81 deg tilt, arrival 5.74 rad/s
  prepend capped from 2.0s to 0.12s (6 frames)
  joint residual 1.15 rad/s
  seam step 0.1843 rad / 0.1746 m / 25.8 rad/s
  against the clip's own 0.1452 rad / 3.88 rad/s
```

The cap traded the joint-limit violation for a violent transition: 2.49 rad in 0.12 s, and a seam angular step 6.7x worse than anything the clip contains between its own frames.

Cause: a cubic's shape is fixed by its endpoints, so duration is the only free parameter, and two demands compete. The mean speed required is the gap over the duration; the arrival speed required is the clip's. When the arrival speed greatly exceeds the mean speed the curve overshoots, and shortening the duration to stop that inflates the mean speed. The balance sits near gap/arrival-speed, about 0.43 s here -- but the gap and the arrival speed differ per joint, so no single duration serves all of them. Taking the minimum across joints lets one demanding joint crush the whole segment.

Fix in progress: scale the arrival velocity per joint so each joint's velocity-time product stays within about 1.5x its own gap, keeping the configured duration. Limits hold, the transition stays gentle, and each joint's seam step is bounded individually rather than one global figure.

Also worth recording: the joint-limit guard added with the cap passed on this clip. It checks reachability, not violence, so it did not catch a legal-but-far-too-fast trajectory. Whatever replaces the cap needs a seam-step bound too, measured against the clip's own frame-to-frame motion, which is the only scale that means anything.

tkevinbest:
Follow up tests and plots. This figure shows the pre-pr behavior and the post-pr behavior. Velocities are actually the derivatives of the positions in the post, and weren't in the pre. 
wbt_kinematics_boxlift

tkevinbest:
Also a simple video showing the results. No appending/prepending is on the far left, the pre-PR behavior is in the middle, and the post PR behavior is on the right. There's minimal visual difference between middle and right, but the velocity references fed to the policy are now correctly differentiated. 

https://github.com/user-attachments/assets/8be411b3-f9ee-4a5d-be01-72e0b527b024

PR #204, opened 2026-09-11
feat(rewards): optionally penalize action rate on the policy mean action

tkevinbest:
## Summary

Adds an opt-in `use_mean_action` flag to `penalty_action_rate` so the action-rate penalty is computed on the policy's deterministic mean action instead of the sampled action. Default behavior is unchanged.

## Problem

`action_rate` is an L2 penalty on the change in action between control steps, used to smooth motion. It reads the action passed to `env.step`, which during training is a sample from the policy's stochastic distribution. Sampling noise adds a roughly constant offset to the penalty, so the policy is charged for exploration noise it cannot control.

For an **unsquashed Gaussian policy (PPO here)** the decomposition is exact:

```
E||a_t - a_{t-1}||^2  =  ||mu_t - mu_{t-1}||^2  +  sum(sigma_t^2) + sum(sigma_{t-1}^2)
```

This identity does **not** apply to FastSAC. Its action is `tanh(N(mu, sigma)) * action_scale + action_bias`; `sigma` lives in pre-tanh space, so the variances do not add this way, and `tanh(mu) * action_scale + action_bias` is the *median* of the action distribution, not `E[a]`. For the squashed case the offset has no closed form, but it does not vanish.

The offset only shrinks as the action std shrinks, which is why this hurts SAC-family algorithms more than PPO:

- **PPO**: std is a state-independent learned parameter that decays over training, so the offset largely goes away on its own. Exception: `min_noise_std` / `min_mean_noise_std` (`agents/modules/ppo_modules.py`) floor it, and when those are configured the offset does not reach zero.
- **FastSAC**: `log_std` is a state-dependent network head clamped to `[log_std_min, log_std_max]`, and entropy autotuning holds it up, so the offset persists and the penalty pulls against the entropy objective. Autotuning is conditional on `config.use_autotune`; with it off, alpha is pinned at `alpha_init`.

Two corroborating observations already in the repo:

- The whole-body-tracking (WBT) FastSAC reward preset uses `action_rate_l2 = -1.0` where the PPO preset uses `-0.1`. A 10x gap is consistent with needing comparable gradient signal through a much larger constant offset.
- Evaluation steps the env with the deterministic mean action, so train-time and eval-time action-rate metrics are not measuring the same quantity.

## Mechanism

The deterministic mean action is threaded alongside the sampled action into the env and stored in `ActionManager` next to the existing sampled-action history.

- `ActionManager` gains `mean_action` / `prev_mean_action` buffers and properties. `process_actions` gains an optional `mean_actions` argument, shape-validated like `actions`. The mean is recorded for reward terms only: it is never split across action terms and never reaches the actuators.
- `BaseTask.step` reads an optional `"mean_actions"` key from `actor_state` and stashes it on the instance rather than widening `_pre_physics_step`. This keeps the `_pre_physics_step(self, actions)` hook signature stable so downstream repos that override it don't break.
- FastSAC `Actor.explore` gains `return_mean`. When set it returns `(action, mean_action)`, where `mean_action` is the squashed-and-scaled value `Actor.forward` already returns as its first output, i.e. the same tensor the inference policy and the ONNX export emit.
- `FastSACEnv.step` gains an optional `mean_actions`; the rollout requests the mean and passes it through. The replay buffer is unchanged, since the mean is only needed at collection time where the reward is computed.
- PPO passes `self.actor.action_mean.detach()`, which it already had on hand for its KL term.
- `penalty_action_rate` gains `use_mean_action: bool = False`. Two byte-identical copies of this term exist, in `managers/reward/terms/wbt.py` and `managers/reward/terms/locomotion.py` (the repo mirrors terms between those modules on purpose); both are updated, and a parametrized test pins that they agree.

When no mean is supplied (inference, replay, `reset_all`), the mean buffers mirror the sampled action, so every existing call site behaves exactly as before.

## Opt-in

`use_mean_action` defaults to `False` and no `config_values/**` preset is touched. Existing runs are unaffected. Consuming repos opt in themselves:

```python
"action_rate_l2": RewardTermCfg(
    func="holosoma.managers.reward.terms.wbt:penalty_action_rate",
    params={"use_mean_action": True},
    weight=-0.1,
),
```

**Opting in changes the scale of the term.** A weight tuned against the noisy penalty will be too strong once the constant offset is gone. The WBT FastSAC `-1.0` likely wants retuning; PPO's `-0.1` is a sensible starting point.

To log both curves side by side, register a second copy of the term with `use_mean_action: True` and a negligible **nonzero** weight (e.g. `-1e-9`). `RewardManager._initialize_terms` skips any term whose `weight == 0.0`, so a zero-weight diagnostic term silently never runs.

## Caveat: the reward is no longer Markov w.r.t. the stored transition

With `use_mean_action=True` the reward depends on `mu(s_t)` and `mu(s_{t-1})`, i.e. on the current policy parameters, not purely on the stored (observation, action) pair. Two consequences:

1. **The critic cannot explain the reward from its inputs.** The `actions` observation term feeds the *sampled* previous action (`managers/observation/terms/wbt.py`, `managers/observation/terms/locomotion.py`), so the previous step's mean is not recoverable from the stored observation. With the default sampled penalty the reward *is* a function of `(obs_t, a_t)`, both of which the critic sees. This is the stronger objection of the two.
2. **Buffered rewards go stale.** Rewards already in the replay buffer were computed with an older mean and drift as the mean moves. On-policy PPO recomputes every iteration, so this doesn't arise there. The usual FastSAC setup (many parallel envs, high update-to-data ratio, short buffer horizon) keeps the drift small, but it is a real approximation.

If this bites, the follow-up is a CAPS-style `lambda * ||mu(s_t) - mu(s_{t+1})||^2` regularizer in the FastSAC actor loss, recomputed fresh each update from the `obs`/`next_obs` pair already in the buffer. Out of scope here.

## Testing

New test files: `managers/reward/tests/test_action_rate_terms.py`, `agents/fast_sac/tests/test_actor_explore.py`.

- `530 passed, 39 skipped` on the full `-m no_sim` suite.
- ruff lint + format clean; mypy reports no new errors on the changed files.
- The new tests pin: the default result is unchanged; `use_mean_action=True` is independent of the sampled action; the mirroring fallback; `reset` for both `env_ids=None` and an index tensor; that a step omitting `"mean_actions"` does not inherit a previously stashed mean; and that `explore`'s returned mean equals both `Actor.forward(...)[0]` and the `deterministic=True` output, for both `use_tanh` values.

GPU verification on an A10G, IsaacSim backend, 128 envs, 2 learning iterations. All exit 0 with checkpoints written:

```
exp:g1-29dof-fast-sac
exp:g1-29dof
exp:g1-29dof-wbt-fast-sac
exp:g1-29dof-wbt
```

That covers the FastSAC and PPO rollout paths for both locomotion and whole-body tracking.

### Measured effect

Both penalty variants were registered in a single whole-body-tracking FastSAC run (3000 env steps, 512 envs) so they see the identical rollout. Raw episode metrics read from tensorboard:

```
overall   sampled=6.92   mean=3.88   noise share=44%   ratio=1.8x
step  100:  42.19 -> 13.25   (69% noise)   # early, entropy high
step 3000:   9.98 ->  6.28   (37% noise)
```

Roughly 35-40% of the penalty this FastSAC policy was paying on this task is exploration noise, and ~69% early in training when entropy is highest.

3000 env steps is early training. This quantifies the noise share, not whether tracking performance improves. A convergence comparison has not been run.


tkevinbest:
No clear resolution on if this was helpful or not. While the sampled action penalizes exploration in addition to mean action change, it has a smoothing effect and acts as a bit of a variance regularizer. There are probably tuning updates we could make to fix this, but for now it is more trouble than it was worth. An interesting result, but not worth continuing.

